PR #250, opened 2018-10-21
Add chat framework for languages

Liareth:
This has a lot of backend changes, but only a few visible client changes:

- Emotes are now coloured (we can change the colour later) with multiple styles. Need to write player documentation on how it works. **, :: ::, [], and "" style emotes work, with a single bracket surround causing that text to come out untranslated (useful for names, e.g. [John]).
- Server chat messages are sent manually.
- Party chat is called comms, and players speaking in it can be overheard by those close to them.

### DIFF
--- a/SWLOR.Game.Server/Event/Module/OnModuleLeave.cs
+++ b/SWLOR.Game.Server/Event/Module/OnModuleLeave.cs
@@ -34,6 +34,11 @@ namespace SWLOR.Game.Server.Event.Module
         {
             NWPlayer pc = (_.GetExitingObject());
 
+            if (pc.IsDM)
+            {
+                App.GetAppState().ConnectedDMs.Remove(pc);
+            }
+
             if (pc.IsPlayer)
             {
                 _.ExportSingleCharacter(pc.Object);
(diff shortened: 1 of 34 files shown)


SWLOR.Game.Server/Event/Module/OnModuleLeave.cs
```diff
@@ -34,6 +34,11 @@ public bool Run(params object[] args)
         {
             NWPlayer pc = (_.GetExitingObject());
 
+            if (pc.IsDM)
+            {
+                App.GetAppState().ConnectedDMs.Remove(pc);
```
zunath: In general I *prefer* to inject the AppState via constructor (just add AppState to constructor, add a member variable and hook it up there). This is essentially using the service locator which is widely considered to be an anti-pattern. The only reason this exists is because I had to break the rules a bit when I removed all of the dependencies out of NWObject.

zunath:
Something is up with Github today. It isn't posting my comments on the code... So here we go, will try again:

Are the changes to the mono API part of stuff you changed in your auto-gen code? Or did you manually change them? Just worried about if we need to regen at some point in the future!

Regarding the App.GetAppState() call:

In general I *prefer* to inject the AppState via constructor (just add AppState to constructor, add a member variable and hook it up there). This is essentially using the service locator which is widely considered to be an anti-pattern. The only reason this exists is because I had to break the rules a bit when I removed all of the dependencies out of NWObject.

PR #600, opened 2018-12-01
Implemented FP command

Couls:
Could remove the _ = script; lines and such but I kept them in just in case they broke something


zunath:
The _ is a reference to INWScript which is what we use to interface with NWN/NWScript. It doesn't hurt anything to leave it there but it's probably less confusing to not inject it unless you need it. 

The only issue I see is "Current" is capitalized unnecessarily. Can you fix that real quick and then I'll merge the change? Other than that, it looks good!

Issue #607, opened 2018-12-01
GAME: Stim Packs useable by anyone


zunath:
Enable all players to use any level of stim packs, without needing the required level or perks. Existing skill checks and perks should enhance their usefulness, but they should not be required.

Reference thread:

https://forums.starwarsnwn.com/t/suggestion-stim-packs/180

Issue #606, opened 2018-12-01
GAME: Repair kits useable by anyone


zunath:
Allow anyone to use repair kits of any type. Existing perks should enhance their usefulness, but not *require* the player have the perk to use them.

Reference thread: https://forums.starwarsnwn.com/t/suggestion-repair-kits/175

PR #612, opened 2018-12-01
Remove Perk Requirements from Repair kits and remove penalty from stimpacks

Couls:
Repair kits:
Each level of perks is a +5 bonus to the durability of the object
Perk description still needs to be changed but that shouldn't be too hard.
I'll add that in once I know this works!
Stimpacks:
Removed attribute penalty from using stimpacks when below level, the only punishment now is gaining reduced XP in medicine

### DIFF
--- a/SWLOR.Game.Server/Data/Migrations/2018-12-02.1.sql
+++ b/SWLOR.Game.Server/Data/Migrations/2018-12-02.1.sql
@@ -1,45 +1,45 @@
 ﻿UPDATE dbo.Perk
-SET Description = 'Provides a bonus to durability when using Electronic repair kits.'
+SET Description = 'Provides a bonus to repairs when using Electronic repair kits.'
 WHERE ID = 152
 UPDATE dbo.Perk
-SET Description = 'Provides a bonus to durability when using armor repair kits'
+SET Description = 'Provides a bonus to repairs when using armor repair kits'
 WHERE ID = 150
 UPDATE dbo.Perk
-SET Description = 'Provides a bonus to durability when using weapon repair kits'
+SET Description = 'Provides a bonus to repairs when using weapon repair kits'
 WHERE ID = 149
 UPDATE dbo.PerkLevel
-SET Description = 'Gains +2 to the durability of armor when using an armor repair kit'
+SET Description = 'Gains +2 to the repair of armor when using an armor repair kit'
 WHERE ID = 2066
 UPDATE dbo.PerkLevel
-SET Description = 'Gains +4 to the durability of armor when using an armor repair kit'
+SET Description = 'Gains +4 to the repair of armor when using an armor repair kit'
 WHERE ID = 2067
 UPDATE dbo.PerkLevel
-SET Description = 'Gains +6 to the durability of armor when using an armor repair kit'
+SET Description = 'Gains +6 to the repair of armor when using an armor repair kit'
 WHERE ID = 2068
 UPDATE dbo.PerkLevel
-SET Description = 'Gains +8 to the durability of armor when using an armor repair kit'
+SET Description = 'Gains +8 to the repair of armor when using an armor repair kit'
 WHERE ID = 2069
 UPDATE dbo.PerkLevel
-SET Description = 'Gains +2 to the durability of armor when using an weapon repair kit'
+SET Description = 'Gains +2 to the repair of weapon when using an weapon repair kit'
 WHERE ID = 2062
 UPDATE dbo.PerkLevel
-SET Description = 'Gains +4 to the durability of armor when using an weapon repair kit'
+SET Description = 'Gains +4 to the repair of weapon when using an weapon repair kit'
 WHERE ID = 2063
 UPDATE dbo.PerkLevel
-SET Description = 'Gains +6 to the durability of armor when using an weapon repair kit'
+SET Description = 'Gains +6 to the repair of weapon when using an weapon repair kit'
 WHERE ID = 2064
 UPDATE dbo.PerkLevel
-SET Description = 'Gains +8 to the durability of armor when using an weapon repair kit'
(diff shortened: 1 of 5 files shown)


Couls:
Should deal with #606 #607

Wandering-Poet:
I imagine you'll want to change the [Perk] description and [PerkLevel] descriptions for the repair kits if they're going to boost the amount of durability they return instead now.

Couls:
> Perk description still needs to be changed but that shouldn't be too hard.
> I'll add that in once I know this works!

SWLOR.Game.Server/Item/RepairKit.cs
```diff
@@ -36,12 +36,11 @@ public void ApplyEffects(NWCreature user, NWItem item, NWObject target, Location
...
             if (skillType == SkillType.Armorsmith)
-                repairAmount += item.CraftBonusArmorsmith;
+                repairAmount += item.CraftBonusArmorsmith + (_perk.GetPCPerkLevel(user.Object,PerkType.ArmorRepair) * 5);
             else if (skillType == SkillType.Weaponsmith)
-                repairAmount += item.CraftBonusWeaponsmith;
+                repairAmount += item.CraftBonusWeaponsmith + (_perk.GetPCPerkLevel(user.Object,PerkType.WeaponRepair) * 5);
             else if (skillType == SkillType.Engineering)
-                repairAmount += item.CraftBonusEngineering;
-
+                repairAmount += item.CraftBonusEngineering + (_perk.GetPCPerkLevel(user.Object,PerkType.ElectronicRepair) * 5);
```
zunath: My first instinct is that these bonuses are way too high. At max, that's I think:

4 * 2 = 8 (tech bonus)
Item bonus (an unknown value between 0-20+)
4 * 5 = 20 (perk bonus)

That's a lot of points! I would feel more comfortable if we started small with 2 and if we need to boost the benefit more, we can. It's just hard to go backwards after we've set the value.

SWLOR.Game.Server/Item/Medicine/StimPack.cs
```diff
@@ -35,17 +35,12 @@ public void ApplyEffects(NWCreature user, NWItem item, NWObject target, Location
             int amount = item.GetLocalInt("AMOUNT") + item.MedicineBonus;
             int rank = player.IsPlayer ? _skill.GetPCSkillRank(player, SkillType.Medicine) : 0;
             int recommendedLevel = item.RecommendedLevel;
-            int delta = recommendedLevel - rank;
-            int penalty = delta / 2;
+           // int delta = recommendedLevel - rank;
+           // int penalty = delta / 2; Remove Penalties from being below recommended level(XP penalty should be enough)
```
zunath: Can you delete these commented lines out?

zunath:
Added some comments on minor issues to fix. Do those and update the DB with a new migration script and we'll be good. Thanks for this!

Couls:
Should be ready to merge now! Check the database to see if the descriptions are to your liking

SWLOR.Game.Server/Data/Migrations/2018-12-02.1.sql
```diff
@@ -0,0 +1,45 @@
...
+WHERE ID = 152
+UPDATE dbo.Perk
+SET Description = 'Provides a bonus to repairs when using armor repair kits'
+WHERE ID = 150
+UPDATE dbo.Perk
+SET Description = 'Provides a bonus to repairs when using weapon repair kits'
+WHERE ID = 149
+UPDATE dbo.PerkLevel
+SET Description = 'Gains +2 to the repair of armor when using an armor repair kit'
+WHERE ID = 2066
```
zunath: Sorry, I should have mentioned some things about DB migrations. In tables where the ID is auto generated (PerkLevel is one of them) you have to do some weird code, since you can't guarantee the ID in your local copy will be the same ID on the server. 

An example of how you could do this is:

UPDATE dbo.PerkLevel
SET Description = 'Gain +2 to armor repair when using an armor repair kit.'
WHERE PerkID =150 AND Level = 1

Perk IDs are manually entered so we're positive they won't change. PerkLevel IDs on the other hand are not, so we have to be explicit. 

This post has some info on finding out whether a column is an IDENTITY: https://stackoverflow.com/questions/87747/how-do-you-determine-what-sql-tables-have-an-identity-column-programmatically

Couls: Ah okay I'll do what you did with the example then

Couls:
Fixed it up

SWLOR.Game.Server/Data/Migrations/2018-12-02.1.sql
```diff
@@ -8,38 +8,38 @@ UPDATE dbo.Perk
...
 UPDATE dbo.PerkLevel
-SET Description = 'Gains +6 to the repair of item when using an electronic repair kit'
-WHERE ID = 1959
+SET Description = 'Gains +6 to electronic repair when using an electronic repair kit'
+WHERE PerkID = 152 AND Level = 3
 UPDATE dbo.PerkLevel
-SET Description = 'Gains +8 to the repair of item when using an electronic repair kit'
-WHERE ID = 1960
+SET Description = 'Gains +8 to electronic repair when using an electronic repair kit'
+WHERE PerkID = 152 AND Level = 1
```
zunath: Should this be Level = 4 ?

Couls: Oh damn miss that

Couls: A bit distracted tonight sorry about that

zunath: NP. I think it looks good now. Thanks!

PR #648, opened 2018-12-03
Implement XP gains from using repair kits

Couls:
I tested it and it seems to work fine. Check through the code to see if everything is to your liking! After this is merged I'll probably be taking a break to study for my finals 😢


SWLOR.Game.Server/Item/RepairKit.cs
```diff
@@ -31,20 +33,47 @@ public CustomData StartUseItem(NWCreature user, NWItem item, NWObject target, Lo
         public void ApplyEffects(NWCreature user, NWItem item, NWObject target, Location targetLocation, CustomData customData)
         {
             SkillType skillType = GetSkillType(item);
+            NWItem targetitem = (target.Object);
             int tech = item.GetLocalInt("TECH_LEVEL");
             float maxDurabilityReductionPenalty = item.GetLocalFloat("MAX_DURABILITY_REDUCTION_PENALTY");
             int repairAmount = tech * 2;
-
+            int skillrank;
+            int Level = targetitem.RecommendedLevel;
```
zunath: Stylistic thing, but make sure variables start with a lower case letter.

SWLOR.Game.Server/Item/RepairKit.cs
```diff
@@ -31,20 +33,47 @@ public CustomData StartUseItem(NWCreature user, NWItem item, NWObject target, Lo
...
+            int delta = 0;
+            int baseXP = 0;
             if (skillType == SkillType.Armorsmith)
-                repairAmount += item.CraftBonusArmorsmith + (_perk.GetPCPerkLevel(user.Object,PerkType.ArmorRepair) * 2);
+            {
+                skillrank = (_skill.GetPCSkillRank((NWPlayer)user, skillType));
```
zunath: Our NW object wrappers suck. To be on the safe side, do user.Object instead of a direct cast. It's one of the things I want to improve. I just haven't had time.

SWLOR.Game.Server/Item/RepairKit.cs
```diff
@@ -31,20 +33,47 @@ public CustomData StartUseItem(NWCreature user, NWItem item, NWObject target, Lo
...
+            else if (delta == 4) baseXP = 325;
+            else if (delta == 3) baseXP = 300;
+            else if (delta == 2) baseXP = 250;
+            else if (delta == 1) baseXP = 225;
+            else if (delta == 0) baseXP = 200;
+            else if (delta == -1) baseXP = 150;
+            else if (delta == -2) baseXP = 100;
+            else if (delta == -3) baseXP = 50;
+            else if (delta == -4) baseXP = 25;
+            _skill.GiveSkillXP((NWPlayer)user,skillType,baseXP,true);
```
zunath: We've reused this case statement a bunch of times. It might make sense for us to put it in a central location, like the SkillService. You don't need to do anything with this, I'm just kind of reminding myself. It's fine how it is at the moment.

Issue #635, opened 2018-12-02
GAME: Weapon debuffs disappear on skill level up


zunath:
When you gain a rank in a weapon skill and you are using a weapon outside of your skill level. (IE Lightsaber skill level 2 and a Lightsaber for Level 12) the debuff you get for using an item outside of your skill range goes away. (Example: Level 2 Lightsaber skill -> -1 to Attack, level up I have +2 to attack now, if I unequip and reequip I get the same debuff as nintended.)	

Jan'ren  Vaal	czs220_maintlvl	49.4754066467285	155.635971069336	-0.490632265806198

PR #654, opened 2018-12-04
Update SkillService.cs

Couls:


Couls:
Just tested on my server this fixes #635

PR #656, opened 2018-12-05
Update PlayerStatService.cs

Couls:
Credit to mith for finding this!


zunath:
What is the impetus for this change?

PR #674, opened 2018-12-09
Add checks for training lightsaber in the finesse perk

Couls:
I assume training lightsabers should be finesseable as well. I believe this additional check should allow for that


Couls:
Closing this and including it in one big PR with other fixes and changes

Issue #671, opened 2018-12-08
GAME: Add option to rename apartment/building


zunath:
Enable players to rename apartments or buildings. There's already code to load this data from the CustomName field. We just need a way to allow players to enter in their own text.

Be sure to check length restrictions on the DB column and incorporate that into the checks in the app code.

Mithreas:
Can I piggy back on this and suggest craftable signs (simple buildings with no interior, renamable)?

zunath:
I've separated your request out into another ticket. https://github.com/zunath/SWLOR_NWN/issues/691

Issue #670, opened 2018-12-08
GAME: Add a way for players to set respawn point to their apartments/houses


zunath:
Allow players to set their respawn point to their buildings or apartments.

Perhaps this could be a structure or maybe just an option when you click on the door.

Issue #585, opened 2018-11-30
GAME: Notify defending owner of tower of an attack


zunath:
If a player attacks a tower, the defending tower's owners (i.e: The person who owns the base and anyone with base permissions) should be notified, if they are online.

PR #693, opened 2018-12-10
Revert CZ-550 to unlinking, Implement new linking behaviour

Couls:
Apply linking behaviour to all creatures outside of CZ-550, Allow renaming and setting residences to respawn points. Warn all players with permissions if their tower is being attacked. Probably some more stuff I missed

### DIFF
--- a/SWLOR.Game.Server/Conversation/BaseManagementTool.cs
+++ b/SWLOR.Game.Server/Conversation/BaseManagementTool.cs
@@ -289,6 +289,10 @@ namespace SWLOR.Game.Server.Conversation
                     SetPageHeader("ConfirmRenamePage", header);
                     ChangePage("ConfirmRenamePage");
                     break;
+                case 2:
+                    GetPC().DeleteLocalInt("LISTENING_FOR_DESCRIPTION");
+                    GetPC().DeleteLocalString("NEW_DESCRIPTION_TO_SET");
+                    break;
             }
         }
         private void HandleConfirmSetNameResponse(int responseID)
@@ -306,6 +310,7 @@ namespace SWLOR.Game.Server.Conversation
                         Guid pcBaseID = new Guid(data.TargetArea.GetLocalString("PC_BASE_ID"));
                         var pcBase = _data.Get<PCBase>(pcBaseID);
                         pcBase.CustomName = GetPC().GetLocalString("NEW_DESCRIPTION_TO_SET");
+                        _data.SubmitDataChange(pcBase, DatabaseActionType.Update);
                         sender.SendMessage("Name is now set to " + pcBase.CustomName);
                     }
                     else if (buildingType == Enumeration.BuildingType.Interior)
@@ -313,6 +318,7 @@ namespace SWLOR.Game.Server.Conversation
                         Guid pcBaseStructureID = new Guid(data.TargetArea.GetLocalString("PC_BASE_STRUCTURE_ID"));
                         var structure = _data.Single<PCBaseStructure>(x => x.ID == pcBaseStructureID);
                         structure.CustomName = GetPC().GetLocalString("NEW_DESCRIPTION_TO_SET");
+                        _data.SubmitDataChange(structure, DatabaseActionType.Update);
                         sender.SendMessage("Name is now set to" + structure.CustomName);
                     }
                     EndConversation();
@@ -839,9 +845,9 @@ namespace SWLOR.Game.Server.Conversation
 
         public override void EndDialog()
         {
-            _base.ClearPlayerTempData(GetPC());
             GetPC().DeleteLocalInt("LISTENING_FOR_DESCRIPTION");
             GetPC().DeleteLocalString("NEW_DESCRIPTION_TO_SET");
+            _base.ClearPlayerTempData(GetPC());     
         }
     }
 }
--- a/SWLOR.Game.Server/Conversation/BaseManagementTool.cs
+++ b/SWLOR.Game.Server/Conversation/BaseManagementTool.cs
@@ -65,7 +65,7 @@ namespace SWLOR.Game.Server.Conversation
                 "75 degrees",
                 "90 degrees",
                 "180 degrees");
-            DialogPage renamePage = new DialogPage("Type a name, Once you are done select confirm.",
+            DialogPage renamePage = new DialogPage("Type a name into the chat box. Once you are done select confirm.",
                 "Confirm");
             DialogPage confirmRenamePage = new DialogPage(
     "<SET LATER>",
--- a/SWLOR.Game.Server/Conversation/BaseManagementTool.cs
+++ b/SWLOR.Game.Server/Conversation/BaseManagementTool.cs
@@ -319,30 +319,6 @@ namespace SWLOR.Game.Server.Conversation
                     break;
             }
         }
-        private void DoRename()
-        {
-            var data = _base.GetPlayerTempData(GetPC());
-            int buildingTypeID = data.TargetArea.GetLocalInt("BUILDING_TYPE");
-            Enumeration.BuildingType buildingType = buildingTypeID <= 0 ? Enumeration.BuildingType.Exterior : (Enumeration.BuildingType)buildingTypeID;
-            data.BuildingType = buildingType;
-            NWPlayer sender = (_.GetPCSpeaker());
-            string text = _.GetPCChatMessage().Trim();
-            _.SetPCChatMessage(string.Empty); // Skip the message
-            _.SendMessageToPC(sender.Object, "New name received. Please press the 'Next' button in the conversation window.");
-                    if (buildingType == Enumeration.BuildingType.Apartment)
-                    {
-                        Guid pcBaseID = new Guid(data.TargetArea.GetLocalString("PC_BASE_ID"));
-                        var pcBase = _data.Get<PCBase>(pcBaseID);
-                        pcBase.CustomName = text;
-                    } 
-                    else if (buildingType == Enumeration.BuildingType.Interior)
-                    {
-                        Guid pcBaseStructureID = new Guid(data.TargetArea.GetLocalString("PC_BASE_STRUCTURE_ID"));
-                        var structure = _data.Single<PCBaseStructure>(x => x.ID == pcBaseStructureID);
-                        structure.CustomName = text;
-                    }
-            _.SendMessageToPC(sender.Object, "New name set to " + text);
-        }
         public override void Back(NWPlayer player, string beforeMovePage, string afterMovePage)
         {
             var data = _base.GetPlayerTempData(GetPC());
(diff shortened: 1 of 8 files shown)


SWLOR.Game.Server/Conversation/BaseManagementTool.cs
```diff
@@ -30,14 +31,16 @@ public class BaseManagementTool : ConversationBase
             IColorTokenService color,
             IDataService data,
             IImpoundService impound,
-            IBasePermissionService perm)
+            IBasePermissionService perm,
+            IPlayerDescriptionService playerdescription)
```
zunath: Picky me here - can you make this _playerDescription and playerDescription?

Couls:
#671 #670 #585

SWLOR.Game.Server/Conversation/BaseManagementTool.cs
```diff
@@ -62,19 +65,26 @@ public override PlayerDialog SetUp(NWPlayer player)
                 "75 degrees",
                 "90 degrees",
                 "180 degrees");
-            
-
+            DialogPage renamePage = new DialogPage("Type a name, Once you are done select confirm.",
```
zunath: Grammar issue here. Maybe use "Type a name into the chat box. Once you are done select confirm."

SWLOR.Game.Server/Conversation/BaseManagementTool.cs
```diff
@@ -62,19 +65,26 @@ public override PlayerDialog SetUp(NWPlayer player)
...
             dialog.AddPage("RetrieveStructurePage", retrievePage);
             dialog.AddPage("RotatePage", rotatePage);
+            dialog.AddPage("RenamePage",renamePage);
+            dialog.AddPage("ConfirmRenamePage", confirmRenamePage);
             return dialog;
         }
 
         public override void Initialize()
         {
+            GetPC().SetLocalInt("LISTENING_FOR_DESCRIPTION", 1);
```
zunath: Confused by this. I don't think it should be listening for a description on initialization of the conversation? This probably needs to be moved to a point when the player selects the rename option.

SWLOR.Game.Server/Conversation/BaseManagementTool.cs
```diff
@@ -250,9 +262,87 @@ public override void DoAction(NWPlayer player, string pageName, int responseID)
...
+                        Guid pcBaseStructureID = new Guid(data.TargetArea.GetLocalString("PC_BASE_STRUCTURE_ID"));
+                        var structure = _data.Single<PCBaseStructure>(x => x.ID == pcBaseStructureID);
+                        structure.CustomName = GetPC().GetLocalString("NEW_DESCRIPTION_TO_SET");
+                        sender.SendMessage("Name is now set to" + structure.CustomName);
+                    }
+                    EndConversation();
+                    break;
+            }
+        }
+        private void DoRename()
```
zunath: Where does this method get called? You shouldn't be able to capture the text entered at this point - you'd have to capture it on the OnChat or OnNWNXChat event. I might be missing something though.

Couls: Ah looks like leftover code

SWLOR.Game.Server/Conversation/EditPrimaryResidence.cs
```diff
@@ -262,7 +285,17 @@ private void DoSetAsResidence()
             ClearNavigationStack();
             ChangePage("MainPage", false);
         }
-
+        private void DoSetRespawnPoint()
```
zunath: Does this actually work inside of the building? I don't think respawn locations can be saved if the player is inside an instance.

Also, does this work for both buildings AND apartments or just buildings? I think there are separate convos for each maybe?

Couls: I've tested with an apartment and it set my respawn position. I may have to do some more testing but I would assume it would work the same way as teleporting you to your instance

SWLOR.Game.Server/Data/Migrations/2018-12-10.2.sql
```diff
@@ -0,0 +1,6 @@
+﻿  UPDATE dbo.SpawnObject
+  SET BehaviourScript = 'PackAggroBehaviour'
+  WHERE NPCGroupID IS NOT NULL
```
zunath: This will modify the behaviours for all NPCs. I don't know that this is correct. There are several different AI groups in the DB: SightAggroBehaviour, SoundAggroRandomWalkBehaviour, SoundAggroBehaviour

You follow up with a fix to the SightAggroBehaviour ones, but the SoundAggroRandomWalkBehaviour entries are still going to get updated.

Couls: I'll go through it and edit the SoundAggro entries back to default

SWLOR.Game.Server/Placeable/ControlTower/OnDamaged.cs
```diff
@@ -50,10 +53,20 @@ public bool Run(params object[] args)
             PCBaseStructure structure = _data.Single<PCBaseStructure>(x => x.ID == new Guid(structureID));
             int maxShieldHP = _base.CalculateMaxShieldHP(structure);
             PCBase pcBase = _data.Get<PCBase>(structure.PCBaseID);
+            var playerIDs = _data.Where<PCBasePermission>(x => x.PCBaseID == structure.PCBaseID).Select(s => s.PlayerID);
+            var toNotify = NWModule.Get().Players.Where(x => playerIDs.Contains(x.GlobalID));
+            bool alerted = false;
```
zunath: This alerted variable seems unnecessary to me. It's always going to be false every time this script runs. Once the script finishes, it'll still be false on the next run.

Moreover, it's going to spam the hell out of players haha. I would look at doing a timestamp (DateTime.UtcNow) and checking against that each time this is fired. If the player hasn't been notified in 15 or 30 seconds (whatever makes sense), then run the following. That way it's immediately apparent there's a problem but they aren't getting spammed about it.

SWLOR.Game.Server/Data/Migrations/2018-12-10.2.sql
Couls: Resolved it by just making it change sightaggros to pack

SWLOR.Game.Server/Placeable/ControlTower/OnDamaged.cs
```diff
@@ -50,10 +53,20 @@ public bool Run(params object[] args)
...
+            if (alerted == false)
+            {
+                foreach(NWPlayer player in toNotify)
+                {
+                    player.SendMessage("One of your bases is under attack!");
```
zunath: I would add the area's name and the sector it's in to this message. Something like:

"Your base in <Area Name> [<Sector>] is under attack!"

Example: Your base in Viscara Wildlands [SW] is under attack!

There are helper methods in the BaseService which should get the sector for you. You'll have to pull back the PCBase from the data service to get the area resref. Then you can either pull from the Area DB entry or just pull it back with NWModule.Get().Areas.Single(x => x.Resref == pcBase.AreaResref)  - Whatever's easiest for you.

SWLOR.Game.Server/Conversation/EditPrimaryResidence.cs
Couls: Just tested and it does work in the building

SWLOR.Game.Server/Conversation/BaseManagementTool.cs
```diff
@@ -250,9 +261,64 @@ public override void DoAction(NWPlayer player, string pageName, int responseID)
...
+                    var data = _base.GetPlayerTempData(GetPC());
+                    int buildingTypeID = data.TargetArea.GetLocalInt("BUILDING_TYPE");
+                    Enumeration.BuildingType buildingType = buildingTypeID <= 0 ? Enumeration.BuildingType.Exterior : (Enumeration.BuildingType)buildingTypeID;
+                    data.BuildingType = buildingType;
+                    NWPlayer sender = (_.GetPCSpeaker());
+                    if (buildingType == Enumeration.BuildingType.Apartment)
+                    {
+                        Guid pcBaseID = new Guid(data.TargetArea.GetLocalString("PC_BASE_ID"));
+                        var pcBase = _data.Get<PCBase>(pcBaseID);
+                        pcBase.CustomName = GetPC().GetLocalString("NEW_DESCRIPTION_TO_SET");
```
zunath: Any time you update a database record (IE: Anything you pull back with _data.Get<>, _data.Single<>, etc. etc. You need to submit a change to the database.

In this case, you need to do _data.SubmitDataChange(pcBase, DatabaseActionType.Update);

Changes to database data is not instant. It gets put into a queue in the background and processed in the order the changes were requested. As long as the server doesn't crash, it's guaranteed to save.

SWLOR.Game.Server/Conversation/BaseManagementTool.cs
```diff
@@ -250,9 +261,64 @@ public override void DoAction(NWPlayer player, string pageName, int responseID)
...
+                        sender.SendMessage("Name is now set to " + pcBase.CustomName);
+                    }
+                    else if (buildingType == Enumeration.BuildingType.Interior)
+                    {
+                        Guid pcBaseStructureID = new Guid(data.TargetArea.GetLocalString("PC_BASE_STRUCTURE_ID"));
+                        var structure = _data.Single<PCBaseStructure>(x => x.ID == pcBaseStructureID);
+                        structure.CustomName = GetPC().GetLocalString("NEW_DESCRIPTION_TO_SET");
```
zunath: You'll want to do a _data.SubmitDataChange(structure, DatabaseActionType.Update); here too

SWLOR.Game.Server/Placeable/ControlTower/OnDamaged.cs
```diff
@@ -50,10 +54,21 @@ public bool Run(params object[] args)
...
+            var playerIDs = _data.Where<PCBasePermission>(x => x.PCBaseID == structure.PCBaseID).Select(s => s.PlayerID);
+            var toNotify = NWModule.Get().Players.Where(x => playerIDs.Contains(x.GlobalID));
+            DateTime timer = DateTime.UtcNow.AddSeconds(30);
+            string clock = timer.ToString(CultureInfo.InvariantCulture);
+            string sector = _base.GetSectorOfLocation(attacker.Location);
+            if (DateTime.UtcNow <= DateTime.Parse(clock))
+            {
+                foreach(NWPlayer player in toNotify)
+                {
+                    player.SendMessage("Your base in " + attacker.Area.Name + " " + sector + "Is under Attack!");
```
zunath: Typo here. "Is under Attack!" should be " is under attack!". Aside from the capital A and I, it will run up against the sector so it would be like "Your base in Area 1 [SW]Is under Attack!"

I'll merge this in and update the text but just wanted to point it out.

PR #695, opened 2018-12-10
Fix wording in renaming

Couls:


Couls:
changes featured in #698

PR #697, opened 2018-12-10
Update DurabilityService.cs

Couls:
capitialize I


zunath:
Are you testing these changes before you submit them? :X

PR #727, opened 2018-12-21
Fix Resuscitation kits

### LINKED ISSUE #718: GAME:  Resuscitation devices drain FP to use
Resuscitation devices drain FP to use

Couls:
recover the targets FP instead of the users

fixes #718

### DIFF
--- a/SWLOR.Game.Server/Item/Medicine/ResuscitationKit.cs
+++ b/SWLOR.Game.Server/Item/Medicine/ResuscitationKit.cs
@@ -84,9 +84,9 @@ namespace SWLOR.Game.Server.Item.Medicine
             baseHeal = (int)(baseHeal * effectivenessPercent);
             if (target.IsPlayer){
             Player dbPlayer = _data.Single<Player>(x => x.ID == target.GlobalID);
+            int fpRecover = (int) (dbPlayer.MaxFP * (0.01f * baseHeal));
             }
             int hpRecover = (int)(target.MaxHP * (0.01f * baseHeal));
-            int fpRecover = (int) (dbPlayer.MaxFP * (0.01f * baseHeal));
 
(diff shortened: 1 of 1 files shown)


SWLOR.Game.Server/Item/Medicine/ResuscitationKit.cs
```diff
@@ -82,10 +82,12 @@ public void ApplyEffects(NWCreature user, NWItem item, NWObject target, Location
...
             baseHeal = (int)(baseHeal * effectivenessPercent);
-
-            Player dbPlayer = _data.Single<Player>(x => x.ID == user.GlobalID);
-            int hpRecover = (int)(target.MaxHP * (0.01f * baseHeal));
+            if (target.IsPlayer)
+            {
+            Player dbPlayer = _data.Single<Player>(x => x.ID == target.GlobalID);
             int fpRecover = (int) (dbPlayer.MaxFP * (0.01f * baseHeal));
+            }
+            int hpRecover = (int)(target.MaxHP * (0.01f * baseHeal));
```
zunath: You've now scoped hpRecover inside of the if statement which is going to cause compilation errors because the later reference to it won't be able to find it.

Can you please make these changes in Visual Studio, confirm it compiles, and test them out before submitting to the PR? Editing them using GitHub is going to lead these types of issues because you can't test them.

Couls: Don't have access to VS at the moment but when I do i'll compile it and test it out but it does look fine, there are no references to fprecover in the code other than there and hprecover is outside of the if statement in my screen

Couls: Ah nevermind I see it now
I'll fix it up properly when I have the time

Issue #725, opened 2018-12-17
GAME: Make power gloves use armorcrafting for mods


zunath:
Power armour gloves are made with Armourcrafting, but require Weapon Modification perks to install mods onto.	V.W Powerglove III	apartment_3	35.1136589050293	24.7788467407227	0

Adjust code to make power gloves use armorcrafting for mod installation.

PR #728, opened 2018-12-21
Powergloves use armorsmith perks

Couls:
#725


SWLOR.Game.Server/Service/ItemService.cs
```diff
@@ -305,10 +305,12 @@ public string OnModuleExamine(string existingDescription, NWPlayer examiner, NWO
         {
             BASE_ITEM_AMULET,
             BASE_ITEM_ARMOR,
+            BASE_ITEM_BRACER,
```
zunath: I'm pretty sure changing this is going to cause problems elsewhere. Gloves are tricky because they're treated as weapons in some places, but armor in others. 

This change will make it so that AC and BAB mods can no longer be installed into any of these items. 

I think I'd recommend setting up a separate static hashset for the armor base items inside of the ModItem class.

Issue #736, opened 2018-12-27
GAME: Add properties to scavenged items, not just harvested ones


Mithreas:
Changes ScavengingPoint code to apply similar logic to Harvesting.
- Updates ResourceService.CalculateChanceForComponentBonus to specify scavenging rather than havesting (determines which skill is used)
- Updates ScavengePoint.OnOpened to use similar logic to ResourceHarvester to apply new properties to the scavenged item
- Updates EffectiveStats to add Scavenging to the list of stats that can be improved by an item.

Mithreas:
PR created: https://github.com/zunath/SWLOR_NWN/pull/737

PR #737, opened 2018-12-27
See https://github.com/zunath/SWLOR_NWN/issues/736 for patch notes.

Mithreas:


.gitignore
```diff
@@ -6,6 +6,7 @@
 *.user
 *.userosscache
 *.sln.docstates
+SWLOR.Game.Server/SWLOR.Game.Server.csproj
```
zunath: I don't know if it's safe to gitignore the csproj file, is it?

SWLOR.Game.Server/ValueObject/EffectiveStats.cs
```diff
@@ -18,6 +18,7 @@ public class EffectiveItemStats
         public int Fabrication { get; set; }
         public int Armorsmith { get; set; }
         public int Harvesting { get; set; }
+        public int Scavenging { get; set; }
```
zunath: You need to add a line to PlayerStatService.GetPlayerItemEffectiveStats to calculate this new property. Otherwise it'll be 0 every time.

SWLOR.Game.Server/Placeable/ScavengePoint/OnOpened.cs
```diff
@@ -106,7 +113,32 @@ public bool Run(params object[] args)
 
                     if (!string.IsNullOrWhiteSpace(spawnItem.Resref) && spawnItem.Quantity > 0)
                     {
-                        _.CreateItemOnObject(spawnItem.Resref, point.Object, spawnItem.Quantity);
+                        NWItem resource = _.CreateItemOnObject(spawnItem.Resref, point.Object, spawnItem.Quantity);
+						
+						// Add properties to the item based on Scavenging skill.  Similar logic to the resource harvester.
+						int ipBonusChance = _resource.CalculateChanceForComponentBonus(oPC, (level/10 + 1), ResourceQuality.Normal, true);
```
zunath: It doesn't seem like there's any restriction on what items get these bonuses. That would mean it's possible for non-components to get component bonuses, I believe. They wouldn't do anything on non-components if that happens because component bonuses are separate properties from the properties which actually do the increases.

May want to do a check on the item and only apply this logic if the item spawned is a component. You should be able to pull this off with something like this:

```

var componentIP = item.ItemProperties.FirstOrDefault(x => _.GetItemPropertyType(x) == (int)CustomItemPropertyType.ComponentType);
if(componentIP != null)
    <apply these changes to the item>

```

SWLOR.Game.Server/SWLOR.Game.Server.csproj
```diff
@@ -1610,7 +1610,7 @@
   </ItemGroup>
   <Import Project="$(MSBuildToolsPath)\Microsoft.CSharp.targets" />
   <PropertyGroup>
-    <PostBuildEvent>xcopy "$(TargetDir)*.*" "C:/NeverwinterNights/NWN/mono" /E /Y /C
+    <PostBuildEvent>xcopy "$(TargetDir)*.*" "C:/NeverwinterNights/github" /E /Y /C
```
zunath: Probably don't want to check in this change for consistency's sake. Or alternatively if you want to set it to the default location on Windows that would be ok too: %userprofile%\documents\Neverwinter Nights\mono should work ok I think.

Mithreas: I have changed this to %userprofile%\documents\Neverwinter Nights (most people do not have a mono folder, unless they are testing locally).

SWLOR.Game.Server/Placeable/ScavengePoint/OnOpened.cs
Mithreas: Applied this filtering.

SWLOR.Game.Server/ValueObject/EffectiveStats.cs
Mithreas: Done - and made some other related changes to make this work end to end.

.gitignore
Mithreas: I've left this change in for now - it essentially means that local changes made by other people won't end up in commits.  Obviously if you need to change the project setup for some reason you'd need to un-Ignore it for that one change.

PR #740, opened 2018-12-28
Plasma Cell && Tranquilizer Interaction

Kenji3108:
Added a quick check on whether tranquilizer is going off or not when applying plasma cell.


Kenji3108:
Did something wrong, my bad!

PR #741, opened 2018-12-28
Plasma Cell && Tranquilizer Interaction

Kenji3108:
Added a quick check on whether tranquilizer is going off or not when applying plasma cell.


SWLOR.Game.Server/Service/AbilityService.cs
```diff
@@ -421,6 +421,7 @@ public void HandlePlasmaCellPerk(NWPlayer player, NWObject target)
             if (_.GetHasFeat((int)CustomFeatType.PlasmaCell, player) == FALSE) return;
             if (player.RightHand.CustomItemType != CustomItemType.BlasterPistol &&
                 player.RightHand.CustomItemType != CustomItemType.BlasterRifle) return;
+            if (target.GetLocalInt("TRANQUILIZER_EFFECT_FIRST_RUN") == NWScript.TRUE) return;
```
zunath: I think this looks right and will work because the tranquilizer effect fires last in the OnModuleApplyDamage method.

Issue #605, opened 2018-12-01
GAME: Plasma Cell toggle


zunath:
Add a way to toggle the effects of Plasma Cell. This is *NOT* a stance, but rather a way to turn the effects of plasma cell on and off. The primary reasoning for turning it off is to avoid attacking mesmerized/slept targets. 

Reference thread: https://forums.starwarsnwn.com/t/suggestion-the-ability-to-toggle-plasma-cell/226

PR #746, opened 2018-12-31
Chat Command: /PlasmaCell

Kenji3108:
This chat command introduces a global variable for the ability to check whether to fire off Plasma Cell damage or not.


Couls:
#605

Issue #722, opened 2018-12-17
GAME: Scavenge on Coxxion Supplies Crates


zunath:
Despite succeedong on 2 scavenging attempts, this box only gives 1 item. This issue occurs continuously with this box even in a new instance. The other boxes don't seem to have this issue	Coxxion Supplies Crate	v_cox_base	36.5297355651855	60.2870330810547	-0.5

zunath:
Despite succeeding on both scavenger checks, this box gave nothing.	Coxxion Supplies Crate	v_cox_base	48.2179260253906	86.9438171386719	0	232.478424072266

Mithreas:
I believe the issue is that nw_it_ld001 exists in both loot table 39 (instance) and 40 (supply crates) but doesn't have a blueprint in the module.  If you know what this is meant to be I can send you a fixed blueprint - or I can fix the database and send you a migration script.  Let me know what you'd rather.

zunath:
I think this was the gold item? I'm not positive, but you probably need to find the gold piece in the standard palette, place it down in the world, right click the object and go to properties to get the resref. 

If the resref is correct then we probably need to look into having a way for loot tables spawn gold.

Mithreas:
Correct resref for gold item is nw_it_gold001

(You can use “show preview window” in the View menu to see it directly from the palette).

I’ll update the live database now and test.  Assuming it works, I will send you a migration script.

Issue #684, opened 2018-12-10
GAME: Containers and +Structure Bonus


zunath:
No furniture or storage container is capable of using +1 structure enhancements or construction parts, only buildings	Alasdaer Quinn	house_int_1	29.9544067382813	26.8325481414795	0

Check and make sure containers actually receive a bonus to +Structure

Couls:
Fixed by Mithreas

PR #748, opened 2019-01-01
Issues684 722

Mithreas:
I realise I didn't make my last change in a branch :(  The only change in this one is the single migration script to address issues #684 and #722.

### DIFF
--- a/SWLOR.Game.Server/Data/Migrations/2019-01-01.1.sql
+++ b/SWLOR.Game.Server/Data/Migrations/2019-01-01.1.sql
@@ -1,3 +1,3 @@
-update swlor.dbo.LootTableItem set Resref='nw_it_gold001' where Resref='nw_it_ld001';
+update dbo.LootTableItem set Resref='nw_it_gold001' where Resref='nw_it_ld001';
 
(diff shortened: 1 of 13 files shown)


SWLOR.Game.Server/Data/Migrations/2019-01-01.1.sql
```diff
@@ -0,0 +1,3 @@
+update dbo.LootTableItem set Resref='nw_it_gold001' where Resref='nw_it_ld001';
```
zunath: Can you double check that this script is marked as "Embedded" in Visual Studio? Right click the file, go to properties, then switch from "Content" and select "Embedded Resource".

Mithreas:
It is definitely embedded in the follow up commit (2f75e66 where I removed the database table names - should also show up in this PR!). 

But I'm not sure whether I need to commit the csproj file to pick that up.  When I embedded it, it didn't seem to change either the file itself or the csproj, so I'm not sure how that gets transmitted to you,

zunath:
I believe it's in the csproj file yeah. If it's in the later commit I'll merge both at the same time.

Issue #749, opened 2019-01-02
GAME: Cannot craft lightsabers with enhancements


Mithreas:
CraftItem.cs calculates the perk to use based on the craft skill.  But now that lightsabers are crafted using their own skill, there's no entry for it - what's needed is one like the below for Medicine. 
`
                    case SkillType.Medicine:
                        model.PlayerPerkLevel = _perk.GetPCPerkLevel(GetPC(), PerkType.MedicalBlueprints);
                        model.EfficiencyLevel = _perk.GetPCPerkLevel(GetPC(), PerkType.MedicineEfficiency);
                        model.OptimizationLevel = _perk.GetPCPerkLevel(GetPC(), PerkType.MedicineOptimization);
                        break;`

PR #750, opened 2019-01-02
Issue749

Mithreas:
Should fix issue #749 - have confirmed code compiles.  Still has my last 3 commits in, haven't figured how to exclude them from future PRs.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,6 +6,7 @@
 *.user
 *.userosscache
 *.sln.docstates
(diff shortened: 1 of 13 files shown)


zunath:
I don't see your other changes in here - just the one which addresses the Lightsaber enhancements. Is it safe to merge this or is something missing?

Mithreas:
It tells me there are 4 commits in it.  If you only see 145c136, then great, that's all you need :)

Issue #586, opened 2018-11-30
GAME: Add ability to replace/upgrade tower


zunath:
Give players an easy way to replace their existing tower with a new one. This should force the tower to have to regenerate shields from zero to prevent abuse. It shouldn't be possible during a siege from other players and it shouldn't let the player to place a tower if they would go over the CPU/Power limits.

Mithreas:
Suggestion: use the new tower on the old one to exchange the structures (if all conditions are met).

PR #762, opened 2019-01-08
Issue586

Mithreas:
Addresses #586.

Tested all the mainline scenarios, including using non-tower items and targeting non-tower targets, all behaving as expected.  

Edge case that the new tower could have more resources in its bay than max.  From a code read this shouldn't cause any harm, so I've not added any code to handle it.


SWLOR.Game.Server/Service/BaseService.cs
```diff
@@ -922,6 +922,96 @@ public int CalculateResourceCapacity(Guid pcBaseID)
...
+            if (string.IsNullOrWhiteSpace(sTowerID)) return "";
+            Guid towerGuid;
+
+            try
+            {
+                towerGuid = new Guid(sTowerID);
+            }
+            catch (Exception e)
+            {
+                Console.Write("Failed to convert GUID: " + sTowerID);
```
zunath: Can you use the ErrorService here so that it'll get pumped into the DB log? This'll let us check the error log just in case someone isn't monitoring the server when it happens

SWLOR.Game.Server/Service/BaseService.cs
```diff
@@ -922,6 +922,96 @@ public int CalculateResourceCapacity(Guid pcBaseID)
...
+            // Actually create/destroy the NWN objects.
+            //--------------------------------------------------------------------------
+            NWArea area = (_.GetAreaFromLocation(_.GetLocation(target)));
+            SpawnStructure(area, towerStructure.ID);
+            target.Destroy();
+            item.Destroy();
+
+            //--------------------------------------------------------------------------
+            // Check whether we have more fuel or resources than we're allowed.
+            //--------------------------------------------------------------------------
```
zunath: This comment is misleading. Suggest either adding the check here or removing the comment.

Mithreas:
Made requested changes.

PR #772, opened 2019-01-10
Issue771

### LINKED ISSUE #771: GAME: Force damage fixes
Missing else statements mean the resistance check logic is broken and could only yield 1/8 or full resist.  Tiers for combat powers are ordered such that higher force stats reduced rather than increasing damage.

Mithreas:
Fixes #771.  Sorry, should have caught the missing else statements when I flagged this code with you yesterday.


zunath:
Sorry about that. My B, my B.

PR #788, opened 2019-01-18
Helmets use "Associated Skill: X Armor" rather than "Custom Item Type…

Mithreas:
…: X Armor".  Add the code to GetSkillTypeForItem to respect this.


zunath:
How come we can't use custom item type? The associated skill thing was meant for something early on, but I'm pretty sure I moved everything over to be based on the item type.

Mithreas:
Well, if you want to update helmets to use the custom item type, go ahead.  But at least the light armour ones in game right now use associated skill type... including ones already crafted.

zunath:
I'll blame Taelon for that one.

PR #855, opened 2019-03-24
Martial Arts Update

Cavcode:
Tried to write a quick fix to allow the Circle Kick feat to actually be applied. Also added in my attempt at scaling damage for unarmed combat.

### DIFF
--- a/SWLOR.Game.Server/Perk/MartialArts/CircleKick.cs
+++ b/SWLOR.Game.Server/Perk/MartialArts/CircleKick.cs
@@ -64,7 +64,7 @@ namespace SWLOR.Game.Server.Perk.MartialArts
 
         public void OnItemEquipped(NWPlayer oPC, NWItem oItem)
         {
-            ApplyFeatChanges(oPC, null);
+            ApplyFeatChanges(oPC, oItem);
         }
 
         public void OnItemUnequipped(NWPlayer oPC, NWItem oItem)
@@ -80,8 +80,11 @@ namespace SWLOR.Game.Server.Perk.MartialArts
         {
             NWItem mainHand = oPC.RightHand;
             NWItem offHand = oPC.LeftHand;
+            NWItem fist = oPC.Arms;
             CustomItemType mainType = mainHand.CustomItemType;
             CustomItemType offType = offHand.CustomItemType;
+            CustomItemType glovesType = fist.CustomItemType;
+
             bool receivesFeat = true;
 
             if (unequippingItem != null && Equals(unequippingItem, mainHand))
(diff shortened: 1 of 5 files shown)


SWLOR.Game.Server/Service/ItemService.cs
```diff
@@ -158,10 +158,6 @@ public string OnModuleExamine(string existingDescription, NWPlayer examiner, NWO
             {
                 description += _color.Orange("Recommended Level: ") + examinedItem.RecommendedLevel + "\n";
             }
-            if (examinedItem.LevelIncrease > 0)
```
zunath: Not sure if this was a merge issue or not but can you add it back in?

SWLOR.Game.Server/SWLOR.Game.Server.csproj
```diff
@@ -2240,7 +2240,6 @@
     <EmbeddedResource Include="Data\Migrations\2019-02-27.3.sql" />
     <EmbeddedResource Include="Data\Migrations\2019-03-02.1.sql" />
     <EmbeddedResource Include="Data\Migrations\2019-03-03.1.sql" />
-    <EmbeddedResource Include="Data\Migrations\2019-03-11.1.sql" />
```
zunath: Shouldn't be removing this file I think. Probably a merge issue too.

SWLOR.Game.Server/Perk/MartialArts/MartialAdept.cs
```diff
@@ -0,0 +1,118 @@
+﻿using NWN;
```
zunath: I think you're missing a database migration script for this perk.

SWLOR.Game.Server/Perk/MartialArts/CircleKick.cs
```diff
@@ -80,8 +80,11 @@ private void ApplyFeatChanges(NWPlayer oPC, NWItem unequippingItem)
         {
             NWItem mainHand = oPC.RightHand;
             NWItem offHand = oPC.LeftHand;
+            NWItem fist = oPC.Arms;
             CustomItemType mainType = mainHand.CustomItemType;
             CustomItemType offType = offHand.CustomItemType;
+            CustomItemType glovesType = fist.CustomItemType;
```
zunath: I see where you're getting the gloves type but don't seem to be doing anything with it? The original code's check was looking for the absence of a weapon. I'm not sure you need to check the glove in this case since we only care about what's equipped?

SWLOR.Game.Server/Perk/MartialArts/CircleKick.cs
```diff
@@ -93,10 +96,12 @@ private void ApplyFeatChanges(NWPlayer oPC, NWItem unequippingItem)
                 offHand = (new Object());
             }
 
-            if ((!mainHand.IsValid && !offHand.IsValid) ||
-                (mainType != CustomItemType.MartialArtWeapon || offType != CustomItemType.MartialArtWeapon))
+            if (mainType != CustomItemType.MartialArtWeapon)
             {
-                receivesFeat = false;
+                if (mainType != CustomItemType.None)
```
zunath: Not every equippable "weapon" will have a custom item type. One example is a torch, should we ever add those. I think you want to stick to the original type of check where you're looking for a martial weapon and the absence of a weapon to maintain existing functionality. Otherwise items which don't fall under our custom item type checks would be treated like an unarmed weapon.

Issue #658, opened 2018-12-06
GAME: Require that a control tower be placed within hours of leasing land


zunath:
There's currently a loophole where players can purchase land but not place a tower down. This leads to land which cannot be claimed by other players via PVP.

My proposed solution at the moment is to require a tower be placed down within 4 hours of purchasing the land (real world time). If a tower doesn't get placed, the land is forfeited and available for purchase by other players. The money paid is not returned.

If land is lost in this way, the player who purchased the land must wait 24 hours (real world time) before purchasing any land again.

Mithreas:
Good idea.  Related suggestions:
- if the PC doesn't have a tower in their inventory, give them an extra "are you sure" dialog.
- If a tower is picked up, start the 4 hour timer again.
- If on module start a property does not have a tower and the timer is not yet underway, start the 4 hour timer.

Couls:
4 hours may be a bit short but with the confirmation dialogue and the fair warning implemented while purchasing it should be fine, i also think you should implement a reminder every 30 minutes that shows how long is left and make it red to put emphasis on The urgency and the fact It’s not refunded

Wandering-Poet:
Sounds reasonable, although for any existing people in the state the timer should start when they log in.

zunath:
Mithreas has put together a solution for this but I think it needs a few tweaks. Also, there's been a ton of refactoring that's happened since he made the PR so we can't do a straight merge. The best course of action here is to take what he did and reimplement it in our current code base.

zunath:
Closing this ticket out as it will no longer be relevant in the revamp.

PR #873, opened 2019-04-04
Issue658

Mithreas:
Addresses #658.

### DIFF
--- a/SWLOR.Game.Server/Conversation/BaseManagementTool.cs
+++ b/SWLOR.Game.Server/Conversation/BaseManagementTool.cs
@@ -690,6 +690,10 @@ namespace SWLOR.Game.Server.Conversation
                     DataService.SubmitDataChange(item, DatabaseActionType.Delete);
                     impoundedCount++;
                 }
+
+                // Reset the purchase time of the base to avoid it being confiscated.
+                pcBase.DateInitialPurchase = DateTime.UtcNow;
+                DataService.SubmitDataChange(pcBase, DatabaseActionType.Update);
             }
             else if (structureType == BaseStructureType.Building)
             {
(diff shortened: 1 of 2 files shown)


SWLOR.Game.Server/Service/BaseService.cs
```diff
@@ -387,7 +388,28 @@ private static void OnModuleHeartbeat()
...
+                    Area dbArea = DataService.Single<Area>(x => x.Resref == pcBase.AreaResref);
+
+                    if (pcBase.DateInitialPurchase < DateTime.UtcNow.AddHours(-6))
+                    {
+                        playerIDs.Add(new Tuple<Guid, string>(pcBase.PlayerID, dbArea.Name + " (" + pcBase.Sector + ")"));
+                        ClearPCBaseByID(pcBase.ID);
+                    }
+                    else
+                    {
+                        players.FirstOrDefault(x => x.GlobalID == pcBase.PlayerID)?.FloatingText("You must put a control tower down within 6 hours of purchase or lose your land in " + dbArea.Name + " (" + pcBase.Sector + ")");
```
zunath: FirstOrDefault runs the chance of returning a null (default = null in this case). Can you swap this to .First() or Single() instead? It probably wouldn't happen in this specific instance, but I want to be on the safe side.

SWLOR.Game.Server/Conversation/BaseManagementTool.cs
```diff
@@ -690,6 +690,10 @@ private void DoRetrieveStructure()
                     DataService.SubmitDataChange(item, DatabaseActionType.Delete);
                     impoundedCount++;
                 }
+
+                // Reset the purchase time of the base to avoid it being confiscated.
+                pcBase.DateInitialPurchase = DateTime.UtcNow;
+                DataService.SubmitDataChange(pcBase, DatabaseActionType.Update);
```
zunath: The only hesitation I have with changing the initial purchase date is that we're actually losing that information by changing it. It's no longer when the player actually bought the land, but when did the timer last update.

I would prefer to have us track the timer in a new column but I'm not married to the idea.

zunath:
We've had a lot of changes to the code base since this was originally written. I still want to implement this, but it's going to require different changes.

Closing this PR for now but we'll use it as a guide as we work on making the adjustments in the new code base.

PR #912, opened 2019-06-08
Implement Helmet and Weapons Appearance Changing

givemedeath:
Feel free to implement as is after testing or tweak the restrictions more. I went through them as thoroughly as I could. I did not remove gaps as iirc selecting a missing weapon or helmet model doesn't do anything terribly bad.


zunath:
Great! Thanks for this

PR #1015, opened 2019-08-25
Adding two behaviours

Cavcode:
Took a shortcut with the waypoint script for now. I translated most of it over, but couldn't quite get it to work the way it was supposed to. Making a call to nw_d2_walkways (a standard NWN resource) which calls the function WalkWaypoints. 

Additionally, added in a bar behaviour that can be expanded upon to get various NPCs to do different emotes / sit in chairs.


Cavcode:
Also changed the stinky womprat quest.

PR #1037, opened 2019-08-30
Various Fixes and Add HoloCom System.

givemedeath:


SWLOR.Game.Server/Conversation/HoloCom.cs
```diff
@@ -0,0 +1,145 @@
...
+
+        public void LoadMainMenu()
+        {            
+            ClearPageResponses("MainPage");
+
+            string header = ColorTokenService.Green("HoloCom Menu\n\n");
+            SetPageHeader("MainPage", header);
+
+            NWPlayer user = GetPC();
+            Console.WriteLine("LoadMainMenu: user = " + user.Name + ", receiver = " + HoloComService.GetCallReceiver(user).Name);
```
zunath: Can you swap these to LogginService calls or remove them? Server's gonna get spammed. If you need them for testing that's ok too. Just need to do something about it before we go live with it.

SWLOR.Game.Server/Service/HoloComService.cs
```diff
@@ -0,0 +1,330 @@
...
+            {
+                ActionPlayAnimation(animation);
+            });
+
+            receiver.DelayAssignCommand(() =>
+            {
+                ActionSpeakString(text, talkvolume);
+            },0.3f);
+
+            /* animation stuff
```
zunath: Probably want to leave this debug info out of the code base.

givemedeath: I left it there so the future change and potential solution is documented somewhere. Should move it to an "Issue" instead?

SWLOR.Game.Server/Conversation/HoloCom.cs
givemedeath: Yeah, i thought i removed that already. Woops

SWLOR.Game.Server/Service/HoloComService.cs
zunath: Yea let's move it to an issue and just tag the PR so we can reference it in the future. I just think it confuses things being in the code itself

PR #1058, opened 2019-09-30
Balance Changes

taelon875:
2019-09-23.1 - Add Grenades to DB
Add Speedy Craftying
--2019-09-23.2.sql
--PerkType.cs
--SpeedyCrafting.cs
--CraftService.cs
(Wanted to Keep only Crafting Perks Still for people that use less crafts.)
PoisonTreatmentkit.cs  - Removes Stat Drains Now and checks for stat drain.
BaseService.s - Discussed With Naia, we feel Silo's Could use a slight boost in efectiveness.
CombatService.cs - This Changes Shield Oath to Provice DI instead of % AC To Further Balancing issues.
Also Properly Adds DI to shields.
LanguageService.cs -  Lowering Language Tick from 5 minutes to 2 Minutes.
PlayerStatService.cs - Limiting AC that gear gives to 12 or 14 for light armor. Then go forward and change Examine to show proper value.
RoleplayService.cs - Doubled Base RP XP Gain, and then added some additonal xp further at the end according to amount RP'ed So values vary and reward mroe for activity.


taelon875:
[agrenade.zip](https://github.com/zunath/SWLOR_NWN/files/3673087/agrenade.zip)
This also goes in with this. Adding the items. for the Grenades. And other item balance stuff. Overwrite everything

taelon875:
Noticed i did not add a confirm statement on the SpeedyCrafting Perk...

SWLOR.Game.Server/Data/Migrations/2019-09-23.2.sql
```diff
@@ -0,0 +1,85 @@
+﻿INSERT INTO dbo.Perk (ID , Name , IsActive , BaseCastingTime , Description , PerkCategoryID , CooldownCategoryID , ExecutionTypeID ,IsTargetSelfOnly ,Enmity , EnmityAdjustmentRuleID , CastAnimationID , ForceBalanceTypeID)
+VALUES	(88 ,
+		N'Speedy Crafting' ,
+		1 ,
+		0 ,
+		N'Reduces the amount of time it takes to Craft. This does not stack with other Speedy Perks' ,
```
zunath: Reword "Craft" to read "craft items" for me. I don't see a reason to keep the individual speedy perks if this one supersedes those, so get rid of the stacking phrasing too.

SWLOR.Game.Server/Enumeration/PerkType.cs
```diff
@@ -90,7 +90,7 @@ public enum PerkType
         ScavengingExpert = 85,
         ForceInsight = 86,
         WeaponFocusLightsaber = 87,
-        // 88 is free
+        SpeedyCrafting = 88,
```
zunath: Let's get rid of the individual speedy crafting perks in this file too. Make sure to add the deletes to the perk info in the DB migration script(s) as well.

SWLOR.Game.Server/Item/Medicine/PoisonTreatmentKit.cs
```diff
@@ -28,7 +28,7 @@ public void ApplyEffects(NWCreature user, NWItem item, NWObject target, Location
                 if (_.GetIsEffectValid(effect) == TRUE)
                 {
                     int effectType = _.GetEffectType(effect);
-                    if (effectType == EFFECT_TYPE_POISON || effectType == EFFECT_TYPE_DISEASE)
+                    if (effectType == EFFECT_TYPE_POISON || effectType == EFFECT_TYPE_DISEASE || effectType == EFFECT_TYPE_ABILITY_DECREASE)
```
zunath: I'm fine with this change. Make sure the item description properly reflects this additional feature.

SWLOR.Game.Server/Service/BaseService.cs
```diff
@@ -1313,7 +1313,7 @@ public static int CalculateMaxFuel(Guid pcBaseID)
                                       if (x == null) return 0;
                                       var baseStructure = DataService.BaseStructure.GetByID(x.BaseStructureID);
                                       return baseStructure.Storage + x.StructureBonus;
-                                  }) * 0.01f;
+                                  }) * 0.1f;
```
zunath: This is a 10% increase now? What do max numbers look like?

SWLOR.Game.Server/Service/BaseService.cs
```diff
@@ -1380,7 +1380,7 @@ public static int CalculateResourceCapacity(Guid pcBaseID)
                                       var baseStructure = DataService.BaseStructure.GetByID(x.BaseStructureID);
 
                                       return baseStructure.Storage + x.StructureBonus;
-                                  }) * 0.01f;
+                                  }) * 0.03f;
```
zunath: Need max numbers on this too please.

SWLOR.Game.Server/Service/CombatService.cs
```diff
@@ -186,8 +186,8 @@ private static void HandleDamageImmunity()
                 int perkLevel = PerkService.GetCreaturePerkLevel(target.Object, PerkType.ShieldProficiency);
                 float perkBonus = 0.02f * perkLevel;
 
-                // DI = 10% + 1% / 3 AC bonuses on the shield + 2% per perk bonus. 
-                reduction = (0.1 + 0.01 * shield.AC / 3) + perkBonus;
+                // DI = 10% + 1% / 3 AC bonuses on the shield + Item AC(Damage Immunity Bonus) + 2% per perk bonus. 
+                reduction = (0.1 + 0.01 * shield.AC / 3) + (shield.GetLocalInt("CUSTOM_ITEM_PROPERTY_AC") * .01) + perkBonus;
```
zunath: Don't access the local ints directly. Use shield.CustomAC instead.

SWLOR.Game.Server/Service/LanguageService.cs
```diff
@@ -127,7 +127,7 @@ public static string TranslateSnippetForListener(NWObject speaker, NWObject list
             lastSkillUp = (lastSkillUp << 32) | (uint)lastSkillUpLow;
             long differenceInSeconds = (now - lastSkillUp) / 10000000;
 
-            if (differenceInSeconds / 60 >= 5)
+            if (differenceInSeconds / 60 >= 2)
```
zunath: This makes it fire once every 2 minutes instead of 5 right? If so, I'm down with this change.

SWLOR.Game.Server/Service/PlayerStatService.cs
```diff
@@ -497,6 +497,12 @@ public static EffectiveItemStats GetPlayerItemEffectiveStats(NWPlayer player, NW
 
                     int itemAC = item.CustomAC;
                     itemAC = CalculateAdjustedValue(itemAC, item.RecommendedLevel, skillRankToUse, 0);
+
+                    if (itemAC > maxAC)
+                    {
+                        //itemAC = maxAC;
+                        item.SetLocalInt("CUSTOM_ITEM_PROPERTY_AC",itemAC = maxAC);
```
zunath: Pull from the item.CustomAC property instead of getting the raw int by variable name.

SWLOR.Game.Server/Service/PlayerStatService.cs
```diff
@@ -473,21 +469,25 @@ public static EffectiveItemStats GetPlayerItemEffectiveStats(NWPlayer player, NW
 
 
                 // Calculate AC
-                if (ItemService.ArmorBaseItemTypes.Contains(item.BaseItemType) ||
-                    ItemService.ShieldBaseItemTypes.Contains(item.BaseItemType))
+                if (ItemService.ArmorBaseItemTypes.Contains(item.BaseItemType))
```
zunath: Why are shields removed from this calculation?

SWLOR.Game.Server/Service/PlayerStatService.cs
```diff
@@ -473,21 +469,25 @@ public static EffectiveItemStats GetPlayerItemEffectiveStats(NWPlayer player, NW
...
                 {
                     int skillRankToUse;
+                    int maxAC = 0;
+
                     if (item.CustomItemType == CustomItemType.HeavyArmor)
                     {
                         skillRankToUse = heavyRank;
+                        maxAC = 12;
```
zunath: Can I get a run down of how you arrived at these caps for each of the armor types?

SWLOR.Game.Server/Service/RoleplayService.cs
```diff
@@ -97,13 +97,13 @@ private static void ProcessPlayerRoleplayXP(NWPlayer player)
             var dbPlayer = DataService.Player.GetByID(player.GlobalID);
             if (dbPlayer.RoleplayPoints >= 50)
             {
-                const int BaseXP = 500;
+                const int BaseXP = 1000;
```
zunath: Why does this need to be doubled?

SWLOR.Game.Server/Service/RoleplayService.cs
```diff
@@ -97,13 +97,13 @@ private static void ProcessPlayerRoleplayXP(NWPlayer player)
...
                 float residencyBonus = PlayerStatService.EffectiveResidencyBonus(player);
                 int xp = (int)(BaseXP + BaseXP * residencyBonus);
                 float dmBonusModifier = dbPlayer.XPBonus * 0.01f;
                 if (dmBonusModifier > 0.25f)
                     dmBonusModifier = 0.25f;
-                xp += (int)(xp * dmBonusModifier);
+                xp += (int)(xp * dmBonusModifier + (dbPlayer.RoleplayPoints * 50));
```
zunath: Purpose of this change?

SWLOR.Game.Server/Data/Migrations/2019-09-23.2.sql
taelon875: I kept the original ones because this line should be more expensive and I don't feel ti is fair to remove what we currently have  to to replace it with a more costly option for the players that only do 1 craft. This line is supposed to help the dedicated crafters that pick up multiple lines.
I can still remove the old ones i just think it would be better to keep them and add some better documentation so players know get one or the other, not both.

SWLOR.Game.Server/Item/Medicine/PoisonTreatmentKit.cs
taelon875: [poisontreatkit.zip](https://github.com/zunath/SWLOR_NWN/files/3682597/poisontreatkit.zip)
Didn't think about that.

SWLOR.Game.Server/Service/BaseService.cs
taelon875: Honestly. Didn't even field test these. 10% may be too much, this is for stronidium. Number are just a shot in the dark.

SWLOR.Game.Server/Service/BaseService.cs
taelon875: This is for resource silo, I think the highest I've gotten on a silo is around 13. So it would be a 39% increase instead of what is currently a 13% increase.

SWLOR.Game.Server/Service/PlayerStatService.cs
taelon875: We had a whole big discussion about the caps, Mith Naia and I all thing 12 is a good start, once we get pushed to test we can see how it changes up stuff though.
The idea is that You can only get +20 Attack Bonus, so we wanted to cap AC a little higher then AB. With 12 per item it gets about 4 more AC then AB or 8 in light armor's case. which should put players back into range for PvP content, hopefully.

SWLOR.Game.Server/Service/RoleplayService.cs
taelon875: Felt the RPXP system wasn't giving at that much in retrospect to what you can get by grinding. Basically 1 kill every 30 minutes via RPXP seems low.

SWLOR.Game.Server/Service/RoleplayService.cs
taelon875: I added this as a way that adds some variability into the numbers you get. Rewarding players who RP more over those that would barely hit the cap as there was none.  It is a flat amount (and now thinking on it that math is wrong is should be  ((dbPlayer.RoleplayPoints - 50) * 50). So you basically get 50 xp for every line over the original trigger. it is added after all other calculations so it doesn't get boosted or anything.

SWLOR.Game.Server/Service/PlayerStatService.cs
taelon875: Basically it makes it so AC is not calculated on shields.  Fix not a change.

SWLOR.Game.Server/Service/LanguageService.cs
taelon875: Yeah, this was decided over just increasing the amounts because this will help the Comprehend Speech Force power as well because you only trigger xp on that when the language triggers for XP.

SWLOR.Game.Server/Service/PlayerStatService.cs
taelon875: Not exactly sure what you mean here...
The reason I am using setlocalint is so the item is set to the right AC on the examine window  just to avoid confusion to players. Otherwise it can say you get 40 ac but you only get 12(or 14).

SWLOR.Game.Server/Data/Migrations/2019-09-23.2.sql
zunath: Can we just lower the SP cost for the universal speedy perk then? They honestly just speed up the crafting process so I'm really fine with everyone picking it up for the same amount that an individual one costs now.

SWLOR.Game.Server/Service/BaseService.cs
zunath: If there's not really a compelling reason for the change, I'd recommend leaving it at 1% per Stronidium Bonus.

SWLOR.Game.Server/Service/BaseService.cs
zunath: I think I'm ok with this one.

SWLOR.Game.Server/Service/PlayerStatService.cs
zunath: There's wrapper code around the CustomAC property which handles conversion of item properties to local ints and some other stuff. It's dangerous to access the variables directly as that conversion code doesn't take place. So instead of doing a SetLocalInt, do item.CustomAC = maxAC;

SWLOR.Game.Server/Service/PlayerStatService.cs
zunath: Ok makes sense. I didn't realize they weren't supposed to grant AC at all :P

SWLOR.Game.Server/Service/PlayerStatService.cs
zunath: I'm good with this.

SWLOR.Game.Server/Service/RoleplayService.cs
zunath: Yea this is adding a minimum of 2500 on every yield. So that's at a minimum of 3500, before XP bonuses are applied. If you swap it to ((dbPlayer.RoleplayPoints - 50) * 50) so that only the excess points count as XP, I can get behind this change.

SWLOR.Game.Server/Service/RoleplayService.cs
zunath: I'm ok with this.

SWLOR.Game.Server/Service/BaseService.cs
zunath: Or alternatively, keep it consistent with the other one at 3% per bonus.

SWLOR.Game.Server/Data/Migrations/2019-09-23.2.sql
taelon875: Sure thing.

PR #1061, opened 2019-10-01
Add Reset Perk Refund Cooldown Command

taelon875:
Adds Perk Refund Cooldown DM Command

### DIFF
--- a/SWLOR.Game.Server/ChatCommand/PerkResetCooldown.cs
+++ b/SWLOR.Game.Server/ChatCommand/PerkResetCooldown.cs
@@ -21,9 +21,8 @@ namespace SWLOR.Game.Server.ChatCommand
                 return;
             }
 
-            var now = DateTime.UtcNow;
             Player dbPlayer = DataService.Player.GetByID(target.GlobalID);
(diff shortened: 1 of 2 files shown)


SWLOR.Game.Server/ChatCommand/PerkResetCooldown.cs
```diff
@@ -0,0 +1,40 @@
...
+        {
+            if (!target.IsPlayer)
+            {
+                user.SendMessage("Only player characters may be targeted with this command.");
+                return;
+            }
+
+            var now = DateTime.UtcNow;
+            Player dbPlayer = DataService.Player.GetByID(target.GlobalID);
+            dbPlayer.DatePerkRefundAvailable = null;
```
zunath: I'd prefer for this to be set to DateTime.UtcNow instead of null.

taelon875: Ahh, is that how you call current time... Okay lemme change that.  (I tried to do that but kept getting issues and Null works XD)

PR #1090, opened 2020-06-28
Fix Dual Blaster bug. Add DM PlayVFX chat command. 

givemedeath:
Give access to entire visualeffects.2da if the effect can be played at a location.

Fix Dual Blaster counting as Throwing.

### DIFF
--- a/SWLOR.Game.Server/ChatCommand/PlayVFX.cs
+++ b/SWLOR.Game.Server/ChatCommand/PlayVFX.cs
@@ -1,0 +1,44 @@
+﻿using SWLOR.Game.Server.ChatCommand.Contracts;
+using SWLOR.Game.Server.Enumeration;
+using SWLOR.Game.Server.GameObject;
+using System;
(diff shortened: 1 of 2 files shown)


SWLOR.Game.Server/ChatCommand/PlayVFX.cs
```diff
@@ -0,0 +1,44 @@
...
+using System;
+using SWLOR.Game.Server.NWN;
+using SWLOR.Game.Server.NWN.Enum;
+using SWLOR.Game.Server.NWN.Enum.VisualEffect;
+using static SWLOR.Game.Server.NWN._;
+using SWLOR.Game.Server.Service;
+
+namespace SWLOR.Game.Server.ChatCommand
+{
+    [CommandDetails("Switches the active language. Use /language help for more information.", CommandPermissionType.DM | CommandPermissionType.Admin)]
```
zunath, line 13: Merged this already but just noticed the description is wrong. Can you fix that? @givemedeath

PR #1110, opened 2020-12-05
Dec update feedback fixes

taelon875:


SWLOR.Game.Server/ChatCommand/GetFP.cs
```diff
@@ -7,7 +7,7 @@
 
 namespace SWLOR.Game.Server.ChatCommand
 {
-    [CommandDetails("Gets the current FP of player.", CommandPermissionType.Player)]
+    [CommandDetails("Gets the current FP of player.", CommandPermissionType.Player | CommandPermissionType.DM | CommandPermissionType.Admin)]
```
zunath, line 10: DMs/Admins don't have DB entries so just adding them to this won't work. Also they don't have FP.

taelon875: no this is making it so DM's can use it on their player characters

SWLOR.Game.Server/ChatCommand/JukeboxDM.cs
```diff
@@ -0,0 +1,24 @@
+﻿using SWLOR.Game.Server.ChatCommand.Contracts;
+using SWLOR.Game.Server.Enumeration;
+using SWLOR.Game.Server.GameObject;
+using SWLOR.Game.Server.Service;
+
+
+namespace SWLOR.Game.Server.ChatCommand
+{
+    [CommandDetails("Opens the Jukebox.", CommandPermissionType.DM | CommandPermissionType.Admin)]
+    public class JukeboxDM : IChatCommand
```
zunath, line 10: This is cool but just call it Jukebox instead of JukeboxDM

SWLOR.Game.Server/ChatCommand/GetFP.cs
zunath: In that case you need to add a check to the method to ensure it doesn't run when they're on their DM characters. It's going to error out because there won't be a record in the DB for them.

SWLOR.Game.Server/ChatCommand/SetScale.cs
```diff
@@ -0,0 +1,35 @@
...
+
+namespace SWLOR.Game.Server.ChatCommand
+{
+    [CommandDetails("Sets Scale.", CommandPermissionType.DM | CommandPermissionType.Admin)]
+    public class SetScale : IChatCommand
+    {
+        public void DoAction(NWPlayer user, NWObject target, NWLocation targetLocation, params string[] args)
+        {
+
+            float value = float.Parse(args[0]);
```
zunath, line 16: You need more error checking on this. If someone does /setScale boogaloo, you're going to get an exception because boogaloo can't be converted to a float.

SWLOR.Game.Server/ChatCommand/SetScale.cs
```diff
@@ -0,0 +1,35 @@
...
+
+            _.SetObjectVisualTransform(target, ObjectVisualTransform.Scale, value);
```
zunath, line 18: Also, this won't be persistent as far as I know. Don't know if that matters or not.

SWLOR.Game.Server/ChatCommand/SetScale.cs
```diff
@@ -0,0 +1,35 @@
...
+
+            user.SendMessage("Object/Person Scale Changed By " + value);
```
zunath, line 20: Nit: Just make the message "Object scale changed to " + value

SWLOR.Game.Server/Perk/ForceAlter/DrainLife.cs
```diff
@@ -78,19 +78,19 @@ public void OnConcentrationTick(NWCreature creature, NWObject target, int spellT
             switch (spellTier)
             {
                 case 1:
-                    amount = 5;
+                    amount = 5 + ((int)ForceBalanceType.Dark - (int)ForceBalanceType.Light);
```
zunath, line 81: This isn't doing what you think it is. You're subtracting two constants (Dark = 2, Light = 1) so this is always going to be (5 + (2 - 1)).

I assume you want to use the player's balance. It's completely slipping my mind how that's calculated and where it's stored but I would check the force resistance code for a place to start.

SWLOR.Game.Server/Service/CombatService.cs
```diff
@@ -470,8 +470,11 @@ public static AbilityResistanceResult CalculateAbilityResistance(NWCreature atta
             //Console.WriteLine("defenderTotal = " + defenderTotal);
             //Console.WriteLine("divisor = " + divisor);
 
-            result.DC = (int) (defenderTotal / divisor * 100);
-            result.Roll = RandomService.D100(1);
+            result.DC = (int)defenderTotal + 10;
```
zunath, line 473: What numbers are coming out of this calculation with these changes? Looks like a 100 point scale dropping to a 20 point scale?

taelon875: The numbers are way more balanced in terms of players hitting each other. instead of all the calculations being done in the DC we split it up so the defender is on the DC and the attacker on the roll. It favors the attackers still but it will even out the wide variation in damages rolled because its not basing the scaled damage on a d100. But it also keeps if the attack is super focused in the force and the other user isn't the low chance of the defender hits correctly.

SWLOR.Game.Server/ChatCommand/GetFP.cs
taelon875: Should be done in the code already ^^

SWLOR.Game.Server/ChatCommand/SetScale.cs
taelon875: Its not an issue atm, DM's should only be using this on NPC's until I can do much further in depth testing, but NPC use is fine.

SWLOR.Game.Server/Perk/ForceAlter/DrainLife.cs
taelon875: Going to revert this for now, I will work on it but I can't get it how I want it then...

SWLOR.Game.Server/Perk/ForceAlter/ForcePush.cs
```diff
@@ -106,6 +106,7 @@ public void OnImpact(NWCreature creature, NWObject target, int perkLevel, int sp
                 }
 
                 _.ApplyEffectToObject(DurationType.Temporary, AbilityService.EffectKnockdown(target, duration), target, duration);
+                _.ApplyEffectToObject(DurationType.Temporary, _.EffectSlow(), target, duration);
```
zunath, line 109: What's the purpose of doing both a knockdown and a slow?

taelon875: That way it still has a effect on people with balance stance up, push is really weird in the fact is incredibly broken or just useless because of balanced stance.

zunath (approved):

PR #1114, opened 2021-01-20
Feature/ffo rewrite

givemedeath:


SWLOR.Game.Server/Feature/AbilityDefinition/ThrowLightsaberAbilityDefinition.cs
```diff
@@ -116,6 +116,18 @@ private static void ThrowLightsaber1(AbilityBuilder builder)
                 .RequirementFP(3)
                 .IsCastedAbility()
                 .DisplaysVisualEffectWhenActivating()
+                .HasCustomValidation((activator, target, level) =>
+                {
+                    var weapon = GetItemInSlot(InventorySlot.RightHand);
```
zunath, line 121: Might be a good idea to throw this logic into a function and reuse it across each level. Less maintenance overhead if it's in one place.

zunath (changes requested):
This looks good other than removing some of the validation duplication on the throw saber levels. You're kicking butt on these! Throw those checks into a method and then I'll merge. Thanks man!

givemedeath:
Requested refactors completed. Also some additional abilities to complete what is in the redesign doc for Force abilities (included light and dark professions). This will not compile until the missing Feat entries are added.

PR #1120, opened 2021-05-03
Combat rules

milliorn:


zunath:
Decided not to do this. I think the restriction isn't necessary any more. Thanks for the contribution just the same though and sorry it took me so long to get back to you on this.

PR #1123, opened 2021-08-28
Fixed Stun and Confusion maybe?

Announcer348:
Changed Stun to measure sphere from correct origin point and aim at correct targets

Increased maximum range of Confusion to not be hugging range, should work properly now but may need further range increase depending; also rearranged if-statement check order for optimization, based on recommendation from professional friend


zunath:
Taelon submitted a PR for additional changes. Closing this one in favor of that.

Issue #1149, opened 2021-10-28
Revamp - Add a flag for visual effects to be applied on creature death


zunath:
Add a mechanism for displaying a visual effect when a creature dies. This value should be configurable via local variable.

The droids in CZ-220 should also be updated to have an explosion effect displayed when they die as part of this ticket.

PR #1177, opened 2021-11-03
Holmstn/feat/vfx on death

HolmstN:
Implements #1149


SWLOR.Web/ClientApp/package-lock.json
```diff
@@ -1,8 +1,17621 @@
 {
```
HolmstN, line 1: hmm this shouldn't be here, I imagine...

SWLOR.Game.Server/Service/SpawnService/SpawnTableBuilder.cs
```diff
@@ -48,13 +48,15 @@ public void RespawnDelay(int minutes)
         /// <param name="type">The object type to spawn</param>
         /// <param name="resref">The resref of the object</param>
         /// <returns>A spawn table builder with the configured settings</returns>
-        public SpawnTableBuilder AddSpawn(ObjectType type, string resref)
+        public SpawnTableBuilder AddSpawn(ObjectType type, string resref, Dictionary<string, int> localVariables = null)
         {
+            Console.WriteLine($"Spawning {resref} with local vars: {string.Join(Environment.NewLine, localVariables)}");
```
HolmstN, line 53: note to self: remove

SWLOR.Game.Server/Service/Creature.cs
```diff
@@ -0,0 +1,28 @@
...
+    public static class Creature
+    {
+        [NWNEventHandler("crea_death")]
+        public static void OnDeath()
+        {
+            var creature = Internal.OBJECT_SELF;
+
+            // apply death VFX
+            var vfx = (VisualEffect)GetLocalInt(creature, "DEATH_VFX_ID");
+            Console.WriteLine(vfx);
```
HolmstN, line 18: note to self: remove

SWLOR.Game.Server/Feature/SpawnDefinition/CZ220SpawnDefinition.cs
```diff
@@ -18,14 +20,22 @@ public class CZ220SpawnDefinition: ISpawnListDefinition
 
         private void DroidSpawns(SpawnTableBuilder builder)
         {
+            var deathAnimation = new Animator()
```
zunath, line 23: This all looks great. The only thing I'm on the fence about is the construction of the animator. It kind of goes against the builder pattern I've set up at this point. I was thinking maybe adding another dot method would be more consistent? My goal with the builders was to make it relatively easy to understand what options are available. 

How do you feel about having .PlayAnimation take in the duration, event, and vfx instead of another object?

I.E:

.PlayAnimation(DurationType.Instant, AnimationEvent.CreateOnDeath, VisualEffect.Fnf_Fireball)

HolmstN: I leaned away from that approach for two reasons:

1) in case we need to extend with new params (Scale already an example), we wouldn't have to modify contracts or write a new overload for the PlayAnimation method

2) If another animation service besides IAnimator was ever used.  Unlikely, but we get the benefit of contract flexibility with the `IAnimator` as the only parameter.

Ultimately these aren't highly persuasive arguments, so I defer to you - let me know if you want me to change to discrete args and I'll code it up.

zunath: Those are fair points, tbh. But if it's all the same to you, let's go ahead and make it a discrete method for now for consistency's sake. Should it expand in scope outside of scale (i.e 5 or more args) we can refactor it to take in the object instead. Basically all you need to do is move the object construction inside the builder. All the stuff you've got now is perfect.

HolmstN: Sounds good to me. Made that change.

zunath (approved):

zunath:
I'm good with all this but noticed it was labeled as a draft still. Can you confirm it's ready to be merged?

HolmstN:
Ready to go, left scale out for now.

HolmstN:
Ready to go!

zunath (approved):

PR #1215, opened 2021-11-21
Shield oath description

Mithreas:
Note - I have made this an embedded resource locally but it doesn't seem to have updated any project files.  Assume this will work fine for others due to the folder settings but shout if you're expecting something else to have changed.


zunath (approved):

zunath:
I needed to mark this script as an Embedded Resource in visual studio (which modifies a value in the csproj file) otherwise the migration process won't pick it up.

No need for you to do anything here, just letting you know if you need to PR in another script change in the future. Thanks for this!

Mithreas:
Yeah, I did that but it didn't seem to save the file to disk.  After I'd committed it and worked on the next fix, the csproj file was updated. Seems there's some sort of caching going on, next time I'll just exit VS before committing to ensure the changes are noted.

zunath:
No worries at all. Easy thing to correct either way. Much appreciated!

PR #1226, opened 2021-11-23
New Tracing Methodology

HolmstN:
Uses OpenTelemtry. Ideally, we should work with NWNX team to export their metrics in OpenTelemetry format instead of hard-dependency on Influx/Grafana.

### TODO:

- segmentation fault
- grafana config


zunath:
We've figured out this issue so I'm going to close this PR out.

PR #1229, opened 2021-11-25
Force VFX Modification

Cavcode:
Added rock, and fireball mrv.
Added kinetic impact vfx.
Added drain effect.
Adjusted drain life, force breach, force crush, force lightning and throw rock to use a vfx and sound.


zunath:
This looks good but holding to merge until an equivalent PR is submitted for revamp.

Issue #1189, opened 2021-11-08
Revamp - Bug report UI


zunath:
Add a UI which appears when the /bug command is used. 

This UI should have fields for the user-entered text. It also should show the player's current location, their CD Key, their account name, and their character name as labels (non editable).

When submitted, the report should fire off to Discord like it does today.

PR #1238, opened 2021-12-04
Implement: Revamp - Bug report UI #1189

givemedeath:


SWLOR.Game.Server/Feature/ChatCommandDefinition/SystemChatCommand.cs
```diff
@@ -180,5 +87,37 @@ private static void ListEmotesCommand(ChatCommandBuilder builder)
...
+                        {
+                            return "You may only use the stuck command every five minutes. Please wait and try again.";
+                        }                        
+                    }
+
+                    return string.Empty;
+                })
+                .Action((user, target, location, args) =>
+                {
+                    var nextStuckAllowed = DateTime.UtcNow.AddMinutes(5);
```
zunath, line 113: Let's increase this to 30 minutes if you don't mind.

SWLOR.Game.Server/Feature/ChatCommandDefinition/SystemChatCommand.cs
```diff
@@ -180,5 +87,37 @@ private static void ListEmotesCommand(ChatCommandBuilder builder)
...
+                    var waypoint = GetNearestObject(Core.NWScript.Enum.ObjectType.Waypoint,user);
```
zunath, line 114: Can you add a using statement to the top so you only need to specify ObjectType.Waypoint here? The namespacing gets hard to read.

zunath (changes requested):

SWLOR.Game.Server/Feature/ChatCommandDefinition/SystemChatCommand.cs
givemedeath: Done

SWLOR.Game.Server/Feature/ChatCommandDefinition/SystemChatCommand.cs
givemedeath: Done

givemedeath:
Added initial implementation of subdual mode.

SWLOR.Game.Server/Feature/ChatCommandDefinition/SystemChatCommand.cs
```diff
@@ -45,105 +44,14 @@ private static void BugCommand(ChatCommandBuilder builder)
                         {
                             return "You may only submit one bug report per minute. Please wait and try again.";
                         }
+                        SetLocalString(user, "BUG_REPORT_LAST_SUBMISSION", dateTime.ToString());
```
zunath, line 47: It might make more sense to do this check when they submit on the UI. As it is now, if they run this command the timer starts as soon as they enter the command. If the user decides to cancel they need to wait a minute before they can open it again. I'd probably get rid of this validation section and do it on the UI's view model for the OnClick event. What do you think?

SWLOR.Game.Server/Feature/GuiDefinition/ViewModel/SettingsViewModel.cs
```diff
@@ -82,6 +92,9 @@ private void UpdateHolonetSetting()
         {
             SetLocalBool(Player, "DISPLAY_HOLONET", DisplayHolonetChannel);
         }
-
+        private void UpdateSubdualMode()
+        {
+            SetLocalBool(Player, "SUBDUAL_MODE", SubdualMode);
```
zunath, line 97: I wouldn't bother with the local variable. The only reason we do it with the Holonet is because of how frequent it is and how many players are involved (would require an indeterminate number of DB calls). I would get rid of this and change your check in Death.cs to check the player DB flag instead.

SWLOR.Game.Server/Service/Death.cs
```diff
@@ -28,10 +28,27 @@ public static void OnPlayerDeath()
                 factionMember = GetNextFactionMember(hostile, false);
             }
 
-            const string RespawnMessage = "You have died. You can wait for another player to revive you or respawn to go to your home point.";
-            PopUpDeathGUIPanel(player, true, true, 0, RespawnMessage);
+            if (GetLocalBool(hostile, "SUBDUAL_MODE"))
```
zunath, line 31: As mentioned in another comment, I'd recommend doing a check here for the DB entity's SubdualMode instead of a local variable.

zunath:
This one's ok but if you don't mind - can you do separate branches and Pull Requests per feature? Getting hard to follow because there are three different tickets happening here. Really appreciate the help by the way!

SWLOR.Game.Server/Feature/ChatCommandDefinition/SystemChatCommand.cs
givemedeath: Good call, it was actually being set in OnClick event already anyway.

givemedeath:
> This one's ok but if you don't mind - can you do separate branches and Pull Requests per feature? Getting hard to follow because there are three different tickets happening here. Really appreciate the help by the way!

Sure, I will try that.

SWLOR.Game.Server/Service/Death.cs
givemedeath: Done

SWLOR.Game.Server/Feature/GuiDefinition/ViewModel/SettingsViewModel.cs
givemedeath: Done

SWLOR.Game.Server/Feature/ChatCommandDefinition/SystemChatCommand.cs
givemedeath: Done

zunath (approved):

PR #1242, opened 2021-12-06
Fuel inventory fix

Mithreas:
As requested.

I've tested this pretty thoroughly with control towers (you'll see some non-obvious code I added in response to testing).  I've not explicitly tested it with starships, but have applied the same fix to the starship computer convo code as to the control tower code.  Would appreciate a second pair of eyes on that comparison.


SWLOR.Game.Server/Event/Module/mod_on_load.cs
```diff
@@ -127,6 +127,8 @@ private static void SetModuleEventScripts()
             NWNXEvents.SubscribeEvent(EventType.DecrementItemStackSizeBefore, "item_dec_stack");
             NWNXEvents.SubscribeEvent(EventType.UseItemBefore, "item_use_before");
             NWNXEvents.SubscribeEvent(EventType.UseItemAfter, "item_use_after");
+            NWNXEvents.SubscribeEvent(EventType.ItemInventoryRemoveItemAfter, "item_inventory_disturb_after");
```
zunath, line 130: I think these script names are too long. There's a cap of 16 characters on resrefs aren't there?

Mithreas: In NWN yes, but since these are all c# classes it seems to work fine.   I've tested and confirmed that this code correctly gets called as-is.

zunath: Interesting. Maybe I'm forgetting how scripts get dispatched in legacy. Anyway, cool! Thanks for the clarification.

zunath (approved):

PR #1244, opened 2021-12-08
Reassembler dialog

Mithreas:
Update the reassembler dialog script to match what the actual reassembler script does.  No change to the function of the reassembler - just makes the description accurate.


SWLOR.Game.Server/Conversation/MolecularReassembly.cs
```diff
@@ -85,18 +85,6 @@ private void LoadSalvagePage()
             // Always create one item with zero bonuses.
             header += componentType.Name + " (No Bonuses) [RL: 0] " + GetChanceColor(100) + "\n";
 
-            // Start by checking attack bonus since we're not storing this value as a local variable on the item.
```
zunath, line 88: I think the fact that this is here and BAB isn't indicates to me that the goal was to remove BAB and replace it with AB. We likely messed up in implementing AB but I feel pretty confident we didn't want BAB to be reassembled. The fact that it still is seems like a bug to me.

Mithreas: I don't feel strongly about this - with revamp coming up my intent wasn't to make any balance changes and just to properly reflect the status quo so it doesn't take reading the code to understand what the tool does.

Originally AB, BAB and AC were all in the reassembler - so I don't think it's quite as simple as 'added AB instead of BAB'.  Ultimately, though, changing this would not have a major impact - it would reduce the BAB possible on bracers by about 1 point [but it will still be possible to get +9 BAB using Martial Arts gloves, so won't reduce max possible BAB] and it would make it possible to make slightly better weapons than we can do today.  So not a major change.

Let me know how you'd rather play this!

zunath (approved):

PR #1246, opened 2021-12-08
Review all perks - first pass

givemedeath:


SWLOR.Game.Server/Feature/AbilityDefinition/Force/ForceMindAbilityDefinition.cs
```diff
@@ -39,10 +39,10 @@ private static void ImpactAction(uint activator, uint target, int level, Locatio
                     break;
             }
             // Damage user.
-            ApplyEffectToObject(DurationType.Instant, EffectDamage((int)(GetCurrentHitPoints(activator) * multiplier)), activator);
+            ApplyEffectToObject(DurationType.Instant, EffectHeal((int)(GetCurrentHitPoints(activator) * multiplier)), activator);
 
             // Recover FP on target.
-            Stat.RestoreFP(activator, (int)(GetCurrentHitPoints(activator) * multiplier));
+            Stat.RestoreFP(activator, -(int)(GetCurrentHitPoints(activator) * multiplier));
```
zunath, line 45: I had forgotten I changed the effects of this perk in the spreadsheet, so good catch.

This line in particular isn't going to work because the RestoreFP call bails out if the value is negative. You'll want to use Stat.ReduceFP instead.

zunath (changes requested):

SWLOR.Game.Server/Feature/AbilityDefinition/Force/ForceMindAbilityDefinition.cs
givemedeath: Done

zunath (approved):

PR #1267, opened 2021-12-16
Feature/resolve attack role

Mithreas:
Implemented and working in mainline cases - not much testing of edge cases done, and the 'defender current target' tracking is still a bit unreliable, but that's only a small circumstance bonus so I wouldn't consider that a shipstopper.

Spotted some issues in the damage plugin I'll work on another day (see comments).  

Main issue identified so far is that Melee Touch attacks (and presumably other sorts of touch attacks) don't appear to pass through the new method.  That will be an issue for force powers when we come to code them.

### DIFF
--- a/SWLOR.Game.Server/Enumeration/AttackType.cs
+++ b/SWLOR.Game.Server/Enumeration/AttackType.cs
@@ -1,0 +1,17 @@
+﻿namespace SWLOR.Game.Server.Enumeration
+{
+// This array is used to signal the preferred attack type of an ability to the native code that manages attack roll overrides.
+// Set the local variable ATTACK_TYPE_OVERRIDE on the attacking creature to one of the valid values below.  This is a one-shot:
+// the value of the variable will be wiped when it is read. 
+//
+// If this override is used, the equipped weapon and any related feats etc. will be ignored.  To make a special weapon that uses
+// (say) Force to attack, build the weapon in the toolset and give it the ATTACK_TYPE_OVERRIDE int variable with the same 
+// values as here.  That weapon will then always make attacks of the specified type, and weapon focus feats etc. will be applied.
+    public enum AttackType
+    {
+        Invalid = 0,
+        Melee = 1,
+        Ranged = 2,
+        Spirit = 3
+    }
+}
--- a/SWLOR.Game.Server/Native/ResolveAttackRoll.cs
+++ b/SWLOR.Game.Server/Native/ResolveAttackRoll.cs
@@ -2,6 +2,8 @@
 using System.Runtime.InteropServices;
 using NWN.Native.API;
 using SWLOR.Game.Server.Core;
+using SWLOR.Game.Server.Enumeration;
+using SWLOR.Game.Server.Service;
 
 namespace SWLOR.Game.Server.Native
 {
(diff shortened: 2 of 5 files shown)


SWLOR.Game.Server/Native/ResolveAttackRoll.cs
```diff
@@ -24,7 +28,366 @@ public static void RegisterHook()
...
+            Log.Write(LogGroup.Attack, "Running OnResolveAttackRoll");
+            var targetObject = CNWSObject.FromPointer(pTarget);
+            if (targetObject == null)
+                return;
+
+            CNWSCreature attacker = CNWSCreature.FromPointer(thisPtr);
+            CNWSCreatureStats attackerStats = attacker.m_pStats;
+            CNWSCombatRound pCombatRound = attacker.m_pcCombatRound;
+
+            Log.Write(LogGroup.Attack, "Attacker: " + attacker.GetFirstName().GetSimple(0) + ", defender " + targetObject.GetFirstName().GetSimple(0));
```
zunath, line 58: The logging you've got in here is fine for now but we probably need to add a way to indicate which logs should be disabled for production use. For example, we'll want all of the Property logs to write regardless of environment but this log information here is going to add a lot of overhead with 40+ people engaging in combat. Only mentioning it here as a reminder, no need for you to make any adjustments yet.

SWLOR.Game.Server/Enumeration/AttackType.cs
```diff
@@ -0,0 +1,17 @@
+﻿namespace SWLOR.Game.Server.Enumeration
+{
+// This array is used to signal the preferred attack type of an ability to the native code that manages attack roll overrides.
+// Set the local variable ATTACK_TYPE_OVERRIDE on the attacking creature to one of the valid values below.  This is a one-shot:
+// the value of the variable will be wiped when it is read. 
+//
+// If this override is used, the equipped weapon and any related feats etc. will be ignored.  To make a special weapon that uses
+// (say) Force to attack, build the weapon in the toolset and give it the ATTACK_TYPE_OVERRIDE int variable with the same
```
zunath, line 8: I think I'd prefer to do this via weapon type, item property, or a combination thereof. The local variables are very hard to see and maintain when you're building stuff. What do you think?

SWLOR.Game.Server/Native/ResolveAttackRoll.cs
```diff
@@ -24,7 +28,366 @@ public static void RegisterHook()
...
+            {
+                defendAttribute = 0;
+            }
+
+            // Defender not targeting the attacker.
+            // Dev note: the GetItem method always creates a new instance of CNWActionNode so there should be no NPEs.
+            // Note: this always returns object invalid for NPCs (2130706432) as their actions aren't represented the same way.
+            uint oidTarget = defender.m_pActionQueue.GetItem(0).oidTarget;
+
+            if (oidTarget == 2130706432) // OBJECT_INVALID
```
zunath, line 154: In this case I think it's safe for you to use NWScript.OBJECT_INVALID instead of hard coding the Id here. It's about the only thing you can safely reference since it's a constant uint value. If it's not possible, can you assign it as a constant in this file by itself?

SWLOR.Game.Server/Enumeration/AttackType.cs
```diff
@@ -0,0 +1,17 @@
...
+//
+// If this override is used, the equipped weapon and any related feats etc. will be ignored.  To make a special weapon that uses
+// (say) Force to attack, build the weapon in the toolset and give it the ATTACK_TYPE_OVERRIDE int variable with the same 
+// values as here.  That weapon will then always make attacks of the specified type, and weapon focus feats etc. will be applied.
+    public enum AttackType
+    {
+        Invalid = 0,
+        Melee = 1,
+        Ranged = 2,
+        Spirit = 3
```
zunath, line 15: I'm curious as to why you went with Spirit instead of Force here? I don't feel particularly strong about it one way or another, just wanted to get your thoughts on it.

SWLOR.Game.Server/Native/ResolveAttackRoll.cs
```diff
@@ -24,7 +28,366 @@ public static void RegisterHook()
...
+                    modifiers -= 10;
+                }
+                else if (range > 20.0f)
+                {
+                    modifiers = -5;
+                }
+                
+            }
+
+            // Attacking from behind.  Does not apply to Force attacks.
```
zunath, line 213: I really like this idea but I'm wondering if it'd be better served as a perk instead of being freely available.

SWLOR.Game.Server/Native/ResolveAttackRoll.cs
```diff
@@ -24,7 +28,366 @@ public static void RegisterHook()
...
+            Log.Write(LogGroup.Attack, "Roll: " + roll.ToString() + ", bonus: " + bonus.ToString());
+            pAttackData.m_nToHitRoll = (byte) (roll / 4);
+            pAttackData.m_nToHitMod = (byte) (bonus / 4);
+            int result = roll - 50 + bonus;
+
+            if (result < 0)
+            { 
+                // Miss
+                pAttackData.m_nAttackResult = 4;
+                pAttackData.m_nMissedBy = (byte) Math.Abs(result); // Dunno if this is needed by anything, but filling it out in case.
```
zunath, line 246: MissedBy would indicate to me that it's the attacker who missed. I assume if you're casting to a byte here it's not a uint type referencing a game object. We should try to track down what this is, if possible. Makes me nervous setting a value but not knowing what it touches

SWLOR.Game.Server/Native/ResolveAttackRoll.cs
Mithreas: Yes, absolutely.  Most of these are debug logging.  I didn't see an option to specify debug vs production - but I figure we'll add that before we launch.  During testing, no reason not to have debug on.

SWLOR.Game.Server/Enumeration/AttackType.cs
Mithreas: Weapon type is by default - you only need the variables if you want to override the default.  Item properties could work for items, but not for abilities - and there's no other way I can see to correspond information from a feat script running in the main engine to the native code.  But if I've missed something shout!

SWLOR.Game.Server/Enumeration/AttackType.cs
Mithreas: Purely because I can see some non-Force abilities using it - e.g. in Medicine you might try to drug someone and want to have that resisted the same way.  I'm just trying to avoid ever wanting to give a mundane character an ability that's tagged 'force'.

SWLOR.Game.Server/Native/ResolveAttackRoll.cs
Mithreas: Could have a perk for extra damage?   
Making this globally available is mainly about giving parties ganging up on someone an advantage - whether that's PCs or NPCs.  It means that even if a single Mynock can't really touch you, a swarm of them will have a decent shot at getting through.  Equally, I don't feel strongly about this if you want to lock it behind a perk.

SWLOR.Game.Server/Native/ResolveAttackRoll.cs
Mithreas: In the original code it's set to 1 if someone rolls a 1, or the difference between the attack roll and the AC if the roll was not 1 and the attack missed.  So I tried to give it equivalent meaning in this code.

It might be used for animation purposes to look for attacks that were near misses?  Or maybe was just created as a hook and never used.  I'd also feel more comfortable knowing what it's for, but not sure how to find out!

SWLOR.Game.Server/Native/ResolveAttackRoll.cs
Mithreas: Good call - should have figured this was already defined somewhere.   Updated (and retested).

SWLOR.Game.Server/Native/ResolveAttackRoll.cs
zunath: Yeah let's keep it. We can investigate adding a perk bonus at a later date.

PR #1268, opened 2021-12-18
PR for today's bug fixes.  Will add commits to it as I figure them out.

Mithreas:


Mithreas:
OK - that's it for the day.  Ready for review.

Module/uti/lightsaber.uti.json
```diff
@@ -129,6 +129,37 @@
           "value": 0
         }
       },
+      {
+        "__struct_id": 0,
```
zunath, line 133: It's always difficult for me to know what's changing in these files. I think you've added some properties to the lightsaber which is probably fine. I just wanted to mention the fact that there is an upgrade process for sabers now. The plan is to enable the player to use them to upgrade their sabers to the next tier instead of fully replacing it. If this has no relevance to the change you made, please disregard.

SWLOR.Game.Server/Feature/AbilityDefinition/FirstAid/MedKitAbilityDefinition.cs
```diff
@@ -48,6 +51,13 @@ private void Impact(uint activator, uint target, int baseAmount)
             var willpowerMod = GetAbilityModifier(AbilityType.Willpower, activator);
             var amount = baseAmount + willpowerMod * 10 + Random.D10(1);
 
+            // limit xp to amount *actually* healed.
```
zunath, line 54: I'm on the fence about this. A big issue I have is with people exploiting direct XP gains. I.E a common one is hurting themselves with one ability and then using Force Heal or an equivalent ability to restore it, so that they get XP.

Generally I'd prefer for the XP gains to be associated with combat using the CombatPoint.AddCombatPointToAllTagged() service method. This ensures the XP gain is consistent with the difficulty of the encounter vs the player's skill. I get that this is an unpopular decision for RP though - e.g why would you only get better at something during combat and not any time you use it?

At a minimum can you add that tagging call here to make sure it links up with the rest of the combat system appropriately? Reach out if you want to discuss approaches to ward off XP abuse too.

SWLOR.Game.Server/Feature/AbilityDefinition/FirstAid/MedKitAbilityDefinition.cs
```diff
@@ -48,6 +51,13 @@ private void Impact(uint activator, uint target, int baseAmount)
...
+            if (amount > (GetMaxHitPoints(target) - GetCurrentHitPoints(target)))
+            {
+                amount = GetMaxHitPoints(target) - GetCurrentHitPoints(target);
+            }
+
+            Service.Skill.GiveSkillXP(activator, SkillType.FirstAid, 3 * amount);
```
zunath, line 60: Nitpick: When possible, default to using statements and remove the namespacing within the code. In this example, removing Service.Skill.GiveSkillXP and replacing it with just Skill.GiveSkillXP.

If there's overlap in class names it's fine to prefix with Service here but I think you can get away without it in this case.

SWLOR.Game.Server/Feature/GuiDefinition/ViewModel/PerksViewModel.cs
```diff
@@ -399,14 +400,14 @@ private void GrantFeats(PerkLevel nextLevel)
             foreach (var feat in nextLevel.GrantedFeats)
             {
                 if (GetHasFeat(feat, Player)) continue;
+                CreaturePlugin.AddFeatByLevel(Player, feat, 1);
```
zunath, line 403: Oops. This was a silly bug. Good catch :)

SWLOR.Game.Server/Feature/ItemDefinition/HarvesterItemDefinition.cs
```diff
@@ -91,11 +93,17 @@ private void Harvester(string tag, int requiredLevel, params string[] validOreTa
                     if (resourceCount <= 0)
                     {
                         DestroyObject(target);
+
+                        // DestroyObject bypasses the OnDeath event, and removes the object so we can't send events.
```
zunath, line 97: I overrode this DestroyObject call to do something for the AI already in Core/Extensions/DestroyObjectExtension.cs

If this is the case, I think I'd prefer to add something in that extension method to look at the OnDeath script and fire it, if it exists. That way we can do it in one place and not have to remember to do it any time we destroy a placeable.

Can you think of any issues with this approach? I know it's outside normal NWN mechanics but it's not intuitive to me the script doesn't fire.

SWLOR.Game.Server/Feature/ScavengePoint.cs
```diff
@@ -113,6 +113,10 @@ public static void OnDisturbed()
                 if (!GetIsObjectValid(firstItem))
                 {
                     DestroyObject(placeable);
+
+                    // DestroyObject bypasses the OnDeath event, and removes the object so we can't send events.
```
zunath, line 117: Re: previous comment - this is what I want to avoid :) Definitely feel strongly about moving the ExecuteScript call in the overridden DestroyObject function after seeing what you've done here.

SWLOR.Game.Server/Feature/StatusEffectDefinition/RestStatusEffectDefinition.cs
```diff
@@ -52,7 +52,7 @@ private void Rest(StatusEffectBuilder builder)
 
                     var hpAmount = 1 + GetAbilityModifier(AbilityType.Vitality, target) / 2;
                     var stmAmount = 1 + GetAbilityModifier(AbilityType.Perception, target) / 2;
-                    var fpAmount = 1 + GetAbilityModifier(AbilityType.Social, target) / 2;
+                    var fpAmount = 1 + GetAbilityModifier(AbilityType.Willpower, target) / 2;
```
zunath, line 55: Good change. I think I did that back when it was still Charisma. This makes a lot more sense.

SWLOR.Game.Server/Service/QuestService/QuestDetail.cs
```diff
@@ -374,6 +374,14 @@ public void Complete(uint player, uint questSource, IQuestReward selectedReward)
             quest.CurrentState = GetStates().Count();
             quest.TimesCompleted++;
 
+            // Note - we must update the database before we give rewards.  Otherwise rewards that update the
```
zunath, line 377: Yet another good catch.

SWLOR.Game.Server/Service/Skill.Mapping.cs
```diff
@@ -59,6 +58,7 @@ private static void LoadItemToSkillMapping()
             _itemToSkillMapping[BaseItem.ShortSpear] = SkillType.TwoHanded;
             _itemToSkillMapping[BaseItem.TwoBladedSword] = SkillType.TwoHanded;
             _itemToSkillMapping[BaseItem.DoubleAxe] = SkillType.TwoHanded;
+            _itemToSkillMapping[BaseItem.Saberstaff] = SkillType.TwoHanded;
```
zunath, line 61: This is a good fix based on the current plan but I wanted to pick your brain about saberstaffs and twin blades. I've been considering moving both of those weapon types to the One-Handed skill given the way NWN mechanics work. I worry it gives one-handed too much weight though, so maybe there are other things we can do to improve it. It also may be a moot point with the attack roll changes you made.

SWLOR.Game.Server/Service/Skill.cs
```diff
@@ -66,7 +66,7 @@ public static void GiveSkillXP(uint player, SkillType skill, int xp)
             {
                 SendMessageToPC(player, $"You earned {details.Name} skill experience. ({xp})");
             }
-            else
+            else if (debtRemoved == 0)
```
zunath, line 69: On line 61 there's a removal of the XP debt but with this change, the player object won't be saved to the DB. I think we need to do that right before the return here to ensure we don't miss that change. I'd recommend doing it here versus after 61 to reduce the number of saves we make during a normal XP gain.

SWLOR.Game.Server/Service/Spawn.cs
```diff
@@ -307,12 +307,18 @@ private static void RemoveQueuedSpawn(QueuedSpawn queuedSpawn)
 
         /// <summary>
         /// When a creature dies, its details need to be queued up for a respawn.
+        /// NOTE: plc_death and crea_death will not trigger if the object is 'killed'
```
zunath, line 310: Just another place that needs attention after we add the script call to the DestroyObject override.

SWLOR.Game.Server/Feature/AbilityDefinition/FirstAid/MedKitAbilityDefinition.cs
Mithreas: Skill conflicted with something else so I needed to qualify it - I forget what.  But I think it was an enum called Skill (maybe in the base NWN stuff, using Skill rather than SkillType?).

SWLOR.Game.Server/Feature/ItemDefinition/HarvesterItemDefinition.cs
Mithreas: The problem with DestroyObject is that it... destroys the object.  So there's nothing left to run an event against (i.e. OBJECT_SELF won't work any more).  OnDeath happens logically *before* DestroyObject, and the object isn't actually destroyed until after all death processing is complete.

I agree my solution here is inelegant.  The best option would be to make EffectDeath() work on Placeables, I think - and then to apply that effect rather than calling DestroyObject.  Actually, checking the Lexicon, it looks like it already does - 'When it is applied to doors or placeable objects, they are destroyed and play a shatter animation.' - so I'll test using that instead.

Module/uti/lightsaber.uti.json
Mithreas: Yes, I added the on hit cast spell power to the lightsaber and saberstaff blueprints.  (I commented that in the commit description but not sure you can easily see that in the review!).

SWLOR.Game.Server/Feature/ItemDefinition/HarvesterItemDefinition.cs
Mithreas: Good news - this worked.  Bad news - it requires toggling the plot flag off, so it's still replacing one line with two lines for placeable destruction.  Commit added.

SWLOR.Game.Server/Service/Skill.cs
Mithreas: I think you might have this the wrong way around - /before/ this change it wasn't being saved to the DB if all our XP was used to pay down debt, because we were returning.  This change means that if we've cleared any XP debt we don't return early, and go through to the save line further down. 

If there's no XP debt and no XP, then we will skip the DB save.

SWLOR.Game.Server/Service/Skill.Mapping.cs
Mithreas: Honestly I would rather have a double weapon skill, since double weapons are interesting to balance for in regular NWN (they can't normally be finessed, but don't get the damage bonus of two handers, so they are usually worse than going dex and using two finessable weapons).  Arelith fixed this by giving an AB bonus to two handed weapons (including double weapons).  On SWLOR, saberstaffs are finessable in Legacy, but double weapons are still in an awkward spot.  (Especially since in Legacy, an offhand weapon stacked with BAB is really strong).

Making double weapons their own skill means we can adjust balance for them uniquely as needed.

Module/uti/lightsaber.uti.json
zunath: I should have checked there, but appreciate the head's up. I'm wondering if we're missing code that applies the OnHit property though? In legacy I was doing that application on equip I believe. Maybe I forgot to carry it over for revamp? I'll add a ticket to review that. It'll be a common thing to forget to add the property so I'd like to automate it.

SWLOR.Game.Server/Feature/AbilityDefinition/FirstAid/MedKitAbilityDefinition.cs
zunath: I think I know what you're talking about. I believe Core has a Skill class too. I've run into that before. All good then, feel free to leave it.

zunath (approved):

PR #1277, opened 2021-12-20
Combat updates

Mithreas:
Lots of files changed, but I suggest you start review with
Combat
GetDamageRoll
ResolveAttackRoll

Everything else should make sense based on those ones.

### DIFF
--- a/SWLOR.Game.Server/Feature/EquipmentStats.cs
+++ b/SWLOR.Game.Server/Feature/EquipmentStats.cs
@@ -94,7 +94,7 @@ namespace SWLOR.Game.Server.Feature
         /// <param name="isAdding">If true, we're adding the HP, if false we're removing it</param>
         private static void ApplyHPBonus(uint player, uint item, ItemProperty ip, bool isAdding)
         {
-            var amount = GetItemPropertyCostTableValue(ip);
+            var amount = GetItemPropertyCostTableValue(ip) * 5;
             var playerId = GetObjectUUID(player);
             var dbPlayer = DB.Get<Player>(playerId);
 
--- a/SWLOR.Game.Server/Feature/PerkDefinition/OneHandedPerkDefinition.cs
+++ b/SWLOR.Game.Server/Feature/PerkDefinition/OneHandedPerkDefinition.cs
@@ -103,19 +103,19 @@ namespace SWLOR.Game.Server.Feature.PerkDefinition
                 .Name("Dual Wield")
 
                 .AddPerkLevel()
-                .Description("Enables the use of two one-handed weapons at the same time at standard NWN penalties.")
+                .Description("Enables the use of two one-handed weapons at the same time at -30%/-50% to hit.")
                 .Price(4)
                 .RequirementSkill(SkillType.OneHanded, 25)
                 .GrantsFeat(FeatType.DualWield)
 
                 .AddPerkLevel()
-                .Description("Grants Two-weapon Fighting feat which reduces attack penalty from -6/-10 to -4/-8 when fighting with two weapons.")
+                .Description("Grants Two-weapon Fighting feat which reduces attack penalty from -30/-50 to -20/-40 when fighting with two weapons.")
                 .Price(5)
                 .RequirementSkill(SkillType.OneHanded, 35)
                 .GrantsFeat(FeatType.TwoWeaponFighting)
 
                 .AddPerkLevel()
-                .Description("Grants Ambidexterity feat which reduces the attack penalty of your off-hand weapon by 4.")
+                .Description("Grants Ambidexterity feat which reduces the attack penalty of your off-hand weapon by 20%.")
                 .Price(6)
                 .RequirementSkill(SkillType.OneHanded, 45)
                 .RequirementCharacterType(CharacterType.Standard)
--- a/SWLOR.Game.Server/Native/GetDamageRoll.cs
+++ b/SWLOR.Game.Server/Native/GetDamageRoll.cs
@@ -4,11 +4,14 @@ using NWN.Native.API;
 using SWLOR.Game.Server.Core;
 using SWLOR.Game.Server.Core.NWScript.Enum;
 using SWLOR.Game.Server.Core.NWScript.Enum.Item;
+using SWLOR.Game.Server.Enumeration;
 using SWLOR.Game.Server.Service;
 using SWLOR.Game.Server.Service.CombatService;
 using BaseItem = SWLOR.Game.Server.Core.NWScript.Enum.Item.BaseItem;
 using EquipmentSlot = NWN.Native.API.EquipmentSlot;
+using InventorySlot = SWLOR.Game.Server.Core.NWScript.Enum.InventorySlot;
 using ObjectType = NWN.Native.API.ObjectType;
+using Random = SWLOR.Game.Server.Service.Random;
 
 namespace SWLOR.Game.Server.Native
 {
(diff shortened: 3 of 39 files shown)


SWLOR.Game.Server/Service/CombatPoint.cs
```diff
@@ -168,14 +140,12 @@ static int CalculateAdjustedXP(int highestRank, int baseXP, SkillType skillType,
...
                     // Armor XP is calculated the same way but is separate from other skills used during combat.
-                    var armorPoints = CalculateArmorPoints(player);
-                    if (armorPoints <= 0) return;
                     var armorRank = dbPlayer.Skills[SkillType.Armor].Rank;
 
                     delta = npcLevel - armorRank;
                     baseXP = Skill.GetDeltaXP(delta);
-                    var xp = CalculateAdjustedXP(armorRank, baseXP, SkillType.Armor, totalPoints, armorPoints, dbPlayer.Skills);
-                    Skill.GiveSkillXP(player, SkillType.Armor, xp);
+                    var xp = baseXP * CalculateRankRangePenalty(highestRank, armorRank);
```
zunath, line 147: I'm thinking maybe we should base the penalty off the NPC level vs the player's armor level instead of tying it back to the weapon. In legacy, we tie it by the highest skill used but I kind of think we should change that. Thoughts?

SWLOR.Game.Server/Service/Item.cs
```diff
@@ -42,6 +43,19 @@ public static void CacheData()
             }
 
             Console.WriteLine($"Loaded {_items.Count} items.");
+
+            // Cache 2da values that we need.  Create a new array for each row, otherwise they
+            // end up pointing to the same array object (and get overwritten).
+            foreach (var baseItem in Enum.GetValues(typeof(BaseItem)).Cast<int>())
+            {
+                string[] values = new string[3];
+                values[0] = Get2DAString("baseitems", "CritThreat", baseItem);
```
zunath, line 52: Can you do the parsing to an int here so it only needs to happen one time rather than on every GetCriticalThreatRange / GetCriticalModifier / GetWeaponSize call?

SWLOR.Game.Server/Service/Spawn.cs
```diff
@@ -318,12 +318,14 @@ public static void QueueRespawn()
             uint creature = OBJECT_SELF;
             var spawnId = GetLocalString(creature, "SPAWN_ID");
             if (string.IsNullOrWhiteSpace(spawnId)) return;
+            if (GetLocalInt(creature, "RESPAWN_QUEUED") == 1) return;
```
zunath, line 321: What does this correct?

SWLOR.Game.Server/Native/GetDamageRoll.cs
```diff
@@ -34,38 +37,85 @@ private static int OnGetDamageRoll(void* thisPtr, void* pTarget, int bOffHand, i
...
+            // subsequent times.  Bail out early.
+            if (targetObject == null) return 0;
 
-            // Calculate attacker's DMG
-            if (attacker != null)
-            {
-                var weapon = bOffHand == 1
+            uint attackType = (uint) AttackType.Melee;
+            // Developer note - the native functions to retrieve a combat round don't work at this point.
+            CNWSItem weapon = bOffHand == 1
```
zunath, line 46: General nitpick: I prefer to use var instead of specific types. Helps with refactoring later. I have a tool that'll automatically change these so it's not a big deal if you don't want to do it now.

zunath (changes requested):

SWLOR.Game.Server/Feature/PerkDefinition/OneHandedPerkDefinition.cs
```diff
@@ -103,19 +103,19 @@ private void DualWield(PerkBuilder builder)
                 .Name("Dual Wield")
 
                 .AddPerkLevel()
-                .Description("Enables the use of two one-handed weapons at the same time at standard NWN penalties.")
+                .Description("Enables the use of two one-handed weapons at the same time at -30%/-50% to hit.")
```
zunath, line 106: I went ahead and updated the spreadsheets to include these changes. I'm using that as our bible so if we end up making mechanical changes to perks, please try to keep the spreadsheet in sync too.

SWLOR.Game.Server/Feature/EquipmentStats.cs
```diff
@@ -94,7 +94,7 @@ public static void RemoveStats()
         /// <param name="isAdding">If true, we're adding the HP, if false we're removing it</param>
         private static void ApplyHPBonus(uint player, uint item, ItemProperty ip, bool isAdding)
         {
-            var amount = GetItemPropertyCostTableValue(ip);
+            var amount = GetItemPropertyCostTableValue(ip) * 5;
```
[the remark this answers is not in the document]
Mithreas: I agree, but I would like to do that closer to release when we've tested these changes more thoroughly.  If the correct multiplier is 3 or 10 or something I would rather not change all items twice.

Also - I think the easier way to do this is to update the display value in the iprp file?  Then people will see whatever we put there rather than needing to update the items themselves.

SWLOR.Game.Server/Service/CombatPoint.cs
Mithreas: baseXP is calculated from the difference between the NPC rank and the armour level, so NPC level is factored in. 

I think you put the rank range penalty in to avoid high level characters levelling lowbies by dragging them around with them?  I'm just trying to preserve that function here, but I'm very happy to remove it.

SWLOR.Game.Server/Service/Item.cs
Mithreas: Sure thing.

SWLOR.Game.Server/Service/Spawn.cs
Mithreas: Sometimes when I scavenged a ScavengePoint it would respawn twice.  Presumably because somehow the destroy event was firing multiple times.  It's general best practice to put this sort of logic around events that are only meant to happen once - stuff firing multiple times seems to be pretty endemic in NWN.

SWLOR.Game.Server/Native/GetDamageRoll.cs
Mithreas: Yeah, I was always taught to be specific about types, was drilled into me when I was learning to code as making code easier for others to understand (unclear how true you find that, but that was how I was trained :)).  So it's a hard habit for me to break :)

SWLOR.Game.Server/Feature/PerkDefinition/OneHandedPerkDefinition.cs
Mithreas: (We'll need to change this again based on our conversation).

Got it - this was really a cosmetic change because the function is the same, I was just mapping the text over to the new mechanics for clarity.  But no reason the same logic doesn't apply to the s/s!

SWLOR.Game.Server/Native/GetDamageRoll.cs
```diff
@@ -34,38 +37,85 @@ private static int OnGetDamageRoll(void* thisPtr, void* pTarget, int bOffHand, i
...
             var damageFlags = attackerStats.m_pBaseCreature.GetDamageFlags();
 
-            var dmg = 0f;
-            var attackAttribute = attackerStats.m_nStrengthBase < 10 ? 0 : attackerStats.m_nStrengthModifier;
-            var damage = 0;
-            var specializationDMGBonus = 0f;
-            Log.Write(LogGroup.Attack, "DAMAGE: attacker strength mod: " + attackAttribute.ToString());
-            // TODO - ranged attacks (and other feat based damage) should not add the strength score.
+            // On a critical hit, this method appears to be invoked multiple times, with an invalid target the second and
+            // subsequent times.  Bail out early.
```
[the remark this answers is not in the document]
Mithreas: I'm not sure, so I've added the check.

SWLOR.Game.Server/Service/CombatPoint.cs
zunath: I think that was the original intention but I'm far less concerned about it now. Let's go ahead and remove it unless you feel strongly about keeping it.

SWLOR.Game.Server/Native/GetDamageRoll.cs
zunath: Haha same here, but I've found it fairly beneficial for the types to be determined by the compiler. On occasion I'll be explicit if it's not clear but in general VS / Resharper works well. Again, just nitpicking - I can live with it either way.

zunath (approved):

PR #1290, opened 2021-12-25
AOO - Auto miss if busy

zunath:
Add section to attack roll to automatically miss if player has an attack of opportunity while busy with another action.

@Mithreas - I wasn't super confident this will address your bug report about using a med kit and getting an AOO. We have a busy flag assigned to the player when they perform an action (refer to Activity service for more info). If the player has this flag and they're performing an AOO, I've set it to automatically miss.

Let me know your thoughts on that as well as the implementation.


Mithreas:
I think you will also want to fill out

zunath:
After testing a bit, I don't think this works. Going to kill the PR and go back to the drawing board.

PR #1295, opened 2021-12-29
Creature updates

Mithreas:
Phew.
* Every creature has been updated and details recorded in the revamp sheet World NPCs tab.  Scripts all checked, and fixed in two cases.
* Attack and damage calculations now respect the Primary Stat and Damage property subtypes.
* Attack subtype is now used for resistance calculations.
* Electrical attacks do double damage to droids (racial type Robot).

New functionality tested but I've hardly begun to test all the new NPCs - that's mainly one for beta testing, though I'll likely crack on a bit myself in Jan.  

Only one restriction.  GetDamageRoll returns just a number, and the engine treats it as physical damage.  I'm not sure whether we have a way to signal damage as a different type back to the engine to display (say) electrical rather than physical.  Not a big deal, but would be more intuitive to players to see (say) magical damage when Force is used, or Acid damage when poison is used.

### DIFF
--- a/Module/utc/byysk_warrior.utc.json
+++ b/Module/utc/byysk_warrior.utc.json
@@ -14,7 +14,7 @@
   },
   "ChallengeRating": {
     "type": "float",
-    "value": 12.0
+    "value": 11.0
   },
   "ClassList": {
     "type": "list",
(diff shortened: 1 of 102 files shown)


zunath:
This is awesome, thanks!

What we might be able to do is ignore any of the default messages the combat system sends using the feedback plugin in nwnx. Then we can do a SendMessageToPC call in the attack and damage rolls with the details we want. Might be worth looking into at a later point.

zunath (approved):

PR #1311, opened 2022-01-06
Module updates

Mithreas:
Various quest and combat system fixes, and CZ-220 updates to add combat training area and fix the bar.

### DIFF
--- a/SWLOR.Game.Server/Native/GetDamageRoll.cs
+++ b/SWLOR.Game.Server/Native/GetDamageRoll.cs
@@ -317,42 +317,54 @@ namespace SWLOR.Game.Server.Native
                     physicalDamage = damage;
                     if (physicalDamage == 0) physicalDamage = 1;
                 }
-                else if (damageType == CombatDamageType.Force && damage > 0)
+                else
                 {
-                    DoFeedback(attacker.m_idSelf, targetObject.m_idSelf, "Force", damage, 255, 255, 255);
-                    ApplyEffectToObject(DurationType.Instant, EffectDamage(damage, DamageType.Divine), targetObject.m_idSelf);
+                    // Assign elemental damage.  We want to do so on a delay because otherwise the damage (and possible target death)
+                    // show up in the combat log before the attack roll (!).  Unfortunately, DelayCommand and AssignCommand->DelayCommand
+                    // don't work here. TODO: find out how to assign/delay commands in native. 
+                    // Alternately, we could find out how to access the damage in a structure like NWNX_Damage and just enter the 
+                    // damage types there directly. 
+                    ProcessElementalDamage(damageType, damage, attacker, targetObject);
                 }
-                else if (damageType == CombatDamageType.Fire && damage > 0)
-                {
-                    DoFeedback(attacker.m_idSelf, targetObject.m_idSelf, "Fire", damage, 255, 0, 0);
-                    ApplyEffectToObject(DurationType.Instant, EffectDamage(damage, DamageType.Fire), targetObject.m_idSelf);
-                }
-                else if (damageType == CombatDamageType.Poison && damage > 0)
-                {
-                    DoFeedback(attacker.m_idSelf, targetObject.m_idSelf, "Poison", damage, 0, 255, 0);
-                    ApplyEffectToObject(DurationType.Instant, EffectDamage(damage, DamageType.Acid), targetObject.m_idSelf);
-                }
-                else if (damageType == CombatDamageType.Electrical && damage > 0)
-                {
-                    DoFeedback(attacker.m_idSelf, targetObject.m_idSelf, "Electrical", damage, 0, 0, 255);
-                    ApplyEffectToObject(DurationType.Instant, EffectDamage(damage, DamageType.Electrical), targetObject.m_idSelf);
-                }
-                else if (damageType == CombatDamageType.Ice && damage > 0)
-                {
-                    DoFeedback(attacker.m_idSelf, targetObject.m_idSelf, "Ice", damage, 0, 255, 255);
-                    ApplyEffectToObject(DurationType.Instant, EffectDamage(damage, DamageType.Cold), targetObject.m_idSelf);
-                }
-
             }
 
             return physicalDamage;
         }
 
+        private static void ProcessElementalDamage(CombatDamageType damageType, int damage, CNWSCreature attacker, CNWSObject targetObject)
+        {
+            if (damageType == CombatDamageType.Force && damage > 0)
+            {
+                DoFeedback(attacker.m_idSelf, targetObject.m_idSelf, "Force", damage, 255, 255, 255);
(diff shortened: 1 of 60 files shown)


SWLOR.Game.Server/Native/GetDamageRoll.cs
```diff
@@ -223,24 +225,51 @@ private static int OnGetDamageRoll(void* thisPtr, void* pTarget, int bOffHand, i
                 if (HasImprovedMultiplier(attacker, weapon)) critMultiplier += 1;
             }
 
+            // Register hit. Do this here in case the target is killed by bonus damage, preventing on hit effects from 
+            // working. This also makes the on hit event unnecessary. 
+            if (!GetIsPC(targetObject.m_idSelf) && !GetIsDM(targetObject.m_idSelf))
```
zunath, line 230: This is a really good idea.

SWLOR.Game.Server/Service/AI.cs
```diff
@@ -326,6 +328,22 @@ private static void LoadAggroEffect()
             ApplyEffectToObject(DurationType.Permanent, effect, OBJECT_SELF);
         }
 
+        private static void DoVFX()
+        {
+            // Allow builders to put permanent effects on creatures - e.g. to make them statues, or make them glow.
+            // Index of standard VFX effects here: https://nwnlexicon.com/index.php?title=Vfx_dur
+            int VFX = GetLocalInt(OBJECT_SELF, "VFX");
```
zunath, line 335: We have a script to do something similar to this, but for placeables. It's in the PlaceableScripts.cs feature. On that one we've got the variable named PERMANENT_VFX_ID

This is probably nit picky but if it's not too much trouble can we use the same variable name here for consistency?

SWLOR.Game.Server/Service/Quest.cs
```diff
@@ -262,7 +263,11 @@ public static void ProgressKillTargetObjectives()
             var possibleQuests = GetQuestsAssociatedWithNPCGroup(npcGroupType);
             if (possibleQuests.Count <= 0) return;
 
-            var killer = GetLastKiller();
+            // We can't use GetLastKiller() as various abilities deal damage that isn't sourced from
+            // the PC.  So get the nearest PC instead.
+            // Ideally we'd use the Combat Point database but that is cleaned up in another method called
+            // by this same handler, so we can't rely on it existing.
+            var killer = GetNearestCreature(CreatureType.PlayerCharacter, 1, creature);
```
zunath, line 270: Are we able to use the enmity service for this or does that also get cleaned up by the time this is run?

SWLOR.Game.Server/Service/CombatPoint.cs
```diff
@@ -24,33 +24,6 @@ public static class CombatPoint
         /// </summary>
         private static readonly Dictionary<uint, HashSet<uint>> _playerToCreatureTracker = new Dictionary<uint, HashSet<uint>>();
 
-        /// <summary>
```
zunath, line 27: This is a good call. There is a method in StandardItemConfiguration.cs which adds the item property in the first place. Its name is ApplyOnHitProperty. Does it make sense to get rid of it as well or do we still have a use for it?

SWLOR.Game.Server/Service/Quest.cs
Mithreas: Same race condition - it's also cleared up in the crea_death event.  If we could order the events we could use it.... is that possible?

SWLOR.Game.Server/Service/CombatPoint.cs
Mithreas: Suggest we leave it for now but kill it before launch if we're sure we don't need to put it back :)

SWLOR.Game.Server/Service/AI.cs
Mithreas: sure thing - let me do that now.

SWLOR.Game.Server/Service/Quest.cs
zunath: The order is not deterministic unfortunately. It's possible to hook the event, do everything, then fire off another script and catch it here. The handful of times I've had to do that it's worked well. I'm just worried the nearest creature might be a different PC than expected.

Mithreas: But they should be in the same party at least, which will make the hook work fine. The odds of having two PCs in different parties both fighting the same creature (and caring about quest kills) don't seem high. 

I'm actually more bothered by this because GetNearestCreature is an expensive method to call and I'd rather not have to :) I was tempted to use the combat point service to tag the last PC to put a combat point on the NPC in a local variable and read that.  It would be a bastardisation of the current system which is all neatly using databases, but using a variable on the NPC means it'll go away when the NPC object is destroyed after death.

SWLOR.Game.Server/Native/GetDamageRoll.cs
```diff
@@ -281,50 +310,61 @@ private static int OnGetDamageRoll(void* thisPtr, void* pTarget, int bOffHand, i
...
 
             return physicalDamage;
         }
 
+        private static void ProcessElementalDamage(CombatDamageType damageType, int damage, CNWSCreature attacker, CNWSObject targetObject)
+        {
+            if (damageType == CombatDamageType.Force && damage > 0)
+            {
+                DoFeedback(attacker.m_idSelf, targetObject.m_idSelf, "Force", damage, 255, 255, 255);
+                ApplyEffectToObject(DurationType.Instant, EffectDamage(damage, DamageType.Divine), targetObject.m_idSelf);
```
zunath: Is it safe to make this call? We're in a native context so I'm worried we don't have direct access to NWScript at this point.

SWLOR.Game.Server/Service/Quest.cs
zunath: I sent you a Discord message, but logging it here too for safekeeping: https://github.com/zunath/SWLOR_NWN/pull/1312

This PR adds "before" and "after" events to the creature events. The idea being that you could hook the "before" events and have access to all the enmity / combat point data as it doesn't get wiped until the "after" event is called. I think this should guarantee you get the right target but I have not tested it yet. Let me know when you get some time and we can look at it.

SWLOR.Game.Server/Native/GetDamageRoll.cs
Mithreas: I replaced this in the 'combine all damage types into a single number' commit.  It's safe to call (as in it works and doesn't crash anything) but dealing damage from here has some side effects around timing that make it awkward.

SWLOR.Game.Server/Service/Quest.cs
Mithreas: Ok - so given the change you've made in the other PR I'll update this to pull from the CombatPoint table and pick something that did damage to the NPC.  When merged with the other PR that should 'just work'.  I won't be able to test until both are merged though,

zunath (approved):

PR #1340, opened 2022-01-16
Combat bug fixes

Mithreas:
- Fixed abusable bug in backstab code
- Fixed damage for twin blades
- Restored the default 0.5 DMG for unarmed strike and weapons without a DMG rating.
- Reduced the range penalty for rifles - they're better at longer range than other ranged weapons.

### DIFF
--- a/SWLOR.Game.Server/Core/Bioware/BiowarePosition.cs
+++ b/SWLOR.Game.Server/Core/Bioware/BiowarePosition.cs
@@ -91,6 +91,9 @@ namespace SWLOR.Game.Server.Core.Bioware
             // X/Y so that we're taking angle relative to the Y axis (X is opposite, Y adjacent)
             var angle = (float)Math.Atan(diffX / diffY);
 
+            // @@@ PROBABLE BUG - this should probably be NWScript.atan not Math.Atan as the latter
+            // returns a result in radians. 
+
             // atan returns -90 to +90.  We need to turn it into a 360 degree facing based on 
             // whether diffX and diffY are positive.  
             // if diffX and diffY are positive, we should have a facing of 0-90.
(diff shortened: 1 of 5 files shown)


SWLOR.Game.Server/Feature/SaveCharacters.cs
```diff
@@ -22,6 +23,13 @@ public static void HandleSaveCharacters()
                     if (GetLocalBool(player, "IS_BARTERING")) continue;
 
                     ExportSingleCharacter(player);
+
+                    // Clear combat state.
+                    if (!GetIsInCombat(player))
```
zunath, line 28: Let's move this to the interval_pc_6s event since it's a completely separate thing from the saving process.

SWLOR.Game.Server/Core/Bioware/BiowarePosition.cs
```diff
@@ -91,6 +91,9 @@ public static float GetRelativeFacing(uint o1, uint o2)
             // X/Y so that we're taking angle relative to the Y axis (X is opposite, Y adjacent)
             var angle = (float)Math.Atan(diffX / diffY);
 
+            // @@@ PROBABLE BUG - this should probably be NWScript.atan not Math.Atan as the latter
```
zunath, line 94: Let's go ahead and switch the Math.Atan call back to the NWScript.atan call if this is causing an issue.

zunath (changes requested):

SWLOR.Game.Server/Core/Bioware/BiowarePosition.cs
Mithreas: It's not causing an issue - but currently nothing calls this method!  Since I have no code set up to test it I didn't want to mess with it today, but I can change it if you like.

SWLOR.Game.Server/Feature/SaveCharacters.cs
Mithreas: Aha!  You already have an event for me to hook into, great.  Let me get that done.

Mithreas: Done.

zunath (approved):

PR #1356, opened 2022-03-02
Final Push - Hultar Changes by Mannax

mannax2012:
[hultarERF_additons_mannax.zip](https://github.com/zunath/SWLOR_NWN/files/8171708/hultarERF_additons_mannax.zip)


zunath:
I've imported the ERF but will not be merging this in as the items/creatures are missing from the erf files. Also ignoring the woods area as it's basically empty and unusable. If there's a newer version of this stuff please let me know.

PR #1418, opened 2022-07-08
Add vfx and sound to space

Cavcode:
**Will require a hak update.**

Added multiple two new blaster bolt projectiles, one green and one red. They are both larger than the normal blaster bolt.

 For each combat module, a sound is played is played via vfx. This allows the sound to be consistently played without having to worry about the action queue.

Added a proton torpedo projectile for the missile launcher.

Added a vfx and sound for when a ship's shield or hull is hit (currently there is no vfx for when a hull is hit, I'll have to make that later on in a later update).

Added UI sounds to be played when targeting or untargetting an enemy.

### DIFF
--- a/SWLOR.Game.Server/Feature/ShipModuleDefinition/CombatLaserModuleDefinition.cs
+++ b/SWLOR.Game.Server/Feature/ShipModuleDefinition/CombatLaserModuleDefinition.cs
@@ -81,6 +81,7 @@ namespace SWLOR.Game.Server.Feature.ShipModuleDefinition
     
                             DelayCommand(0.3f, () =>
                             {
+                                PlayShipHitSfx(target, damage);
                                 Space.ApplyShipDamage(activator, target, damage);
                             });
                         });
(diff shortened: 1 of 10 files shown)


SWLOR.Game.Server/Feature/ShipModuleDefinition/CombatLaserModuleDefinition.cs
```diff
@@ -99,5 +103,31 @@ public class CombatLaserModuleDefinition: IShipModuleListDefinition
...
+            if (amount < 0) return;
+
+            var targetShipStatus = Space.GetShipStatus(target);
+
+            var remainingDamage = amount;
+            // First deal damage to target's shields.
+            if (remainingDamage <= targetShipStatus.Shield)
+            {
+                // Shields have enough to cover the attack.
+                targetShipStatus.Shield -= remainingDamage;
```
zunath, line 117: Space.ApplyShipDamage on line 84 is already doing this. This will apply damage twice. This section should be removed. 

I might be ok with those two VFX going into Space.ApplyShipDamage if you want to keep it though.

zunath (approved):

PR #1456, opened 2022-07-28
Implement XP/Combat Point changes

### LINKED ISSUE #1446: XP/Combat Point Changes
1. Give an XP group classification to each skill. One handed, two handed, ranged, martial arts would be under one group. Force and devices in another.
2. XP would share within the group rather than across all skills. Force/devices are unaffected since you can only use one or the other on a single character.
3. Remove the rank delta check.

JohnWildkins:
Closes #1446 

This PR prevents over-leveled Force / Devices from affecting weapon skill XP gain and vice versa; it also exempts First Aid from affecting either category. The First Aid `Med Kit` ability now no longer directly gives XP for use, unless it is explicitly used out-of-combat up to two minutes after an engagement. It will still provide combat points -> XP on kill when used in combat.

Weapon Skills: One-Handed, Two-Handed, Martial Arts, Ranged
Utility: Force, Devices
Exempt: First Aid, Armor

XP gain is still based on highest skill level of its category vs. the enemy fought, so using a level 50 One-Handed weapon and a level 5 Two-Handed weapon against a level 20 enemy will still result in no XP gained for weapons; the difference now is that Force, Devices, and First Aid XP will be unaffected.

### DIFF
--- a/SWLOR.Game.Server/Service/CombatPoint.cs
+++ b/SWLOR.Game.Server/Service/CombatPoint.cs
@@ -109,37 +109,56 @@ namespace SWLOR.Game.Server.Service
                         .Where(x => cpList.ContainsKey(x.Key))
                         .ToDictionary(x => x.Key, y => y.Value);
 
-                    var highestRank = skillsWithCP
-                        .Where(x => x.Key != SkillType.Armor)
-                        .OrderByDescending(o => o.Value.Rank)
-                        .Select(s => s.Value.Rank)
-                        .First();
-
-                    // Base amount of XP is determined by the player's highest-leveled skill rank versus the creature's level.
-                    var delta = npcLevel - highestRank;
-                    var baseXP = Skill.GetDeltaXP(delta);
-                    var totalPoints = (float)cpList.Sum(s => s.Value);
                     var areaBonus = GetLocalInt(GetArea(npc), "AREA_XP_BONUS_PERCENTAGE") * 0.01f;
 
-                    // Each skill used during combat receives a percentage of the baseXP amount depending on the number of combat points earned.
-                    foreach (var (skillType, cp) in cpList)
+                    foreach (CombatPointCategory cpCategory in Enum.GetValues(typeof(CombatPointCategory)))
                     {
-                        var percentage = cp / totalPoints;
-                        var adjustedXP = baseXP * percentage;
-                        adjustedXP += areaBonus * adjustedXP;
+                        var validSkills = skillsWithCP
+                            .Where(x => Skill.GetSkillDetails(x.Key).XPType == cpCategory)
+                            .ToDictionary(x => x.Key, y => y.Value);
 
-                        Skill.GiveSkillXP(player, skillType, (int)adjustedXP);
-                    }
+                        // Base amount of XP is determined by the player's highest-leveled skill rank in each XP category versus the creature's level.
+                        if (cpCategory != CombatPointCategory.Exempt)
+                        {
+                            var validCPs = cpList
+                                .Where(x => Skill.GetSkillDetails(x.Key).XPType == cpCategory);
+                            if (!validCPs.Any()) continue;
+                            if (!validSkills.Any()) continue;
 
-                    // Armor XP is calculated the same way but is separate from other skills used during combat.
-                    if (!Space.IsPlayerInSpaceMode(player))
-                    {
-                        var armorRank = dbPlayer.Skills[SkillType.Armor].Rank;
-                        delta = npcLevel - armorRank;
-                        baseXP = Skill.GetDeltaXP(delta);
-                        var adjustedXP = baseXP + areaBonus * baseXP;
+                            var highestRank = validSkills
+                                .OrderByDescending(o => o.Value.Rank)
+                                .Select(s => s.Value.Rank)
+                                .First();
 
-                        Skill.GiveSkillXP(player, SkillType.Armor, (int)adjustedXP);
+                            var xpDelta = npcLevel - highestRank;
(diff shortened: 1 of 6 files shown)


SWLOR.Game.Server/Service/CombatPoint.cs
```diff
@@ -276,24 +295,26 @@ public static void ClearRecentEnemyLevel(uint player)
         /// <param name="player">The player to receiving the point.</param>
         /// <param name="skill">The skill to associate with the point.</param>
         /// <param name="amount">The number of points to add.</param>
-        /// <returns>True if at least one creature is tagged, false otherwise.</returns>
-        public static void AddCombatPointToAllTagged(uint player, SkillType skill, int amount = 1)
+        /// <returns>Number of creatures tagged by the player, -1 if the tracker has expired or is unavailable.</returns>
+        public static int AddCombatPointToAllTagged(uint player, SkillType skill, int amount = 1)
```
zunath, line 299: I think we're violating single responsibility here. I would suggest keeping this as void and instead adding a public method to retrieve the number of creatures tagged by the player. This will result in fewer dependencies should we need to refactor this method in the future.

SWLOR.Game.Server/Service/SkillService/SkillCategoryType.cs
```diff
@@ -16,6 +16,13 @@ public enum SkillCategoryType
         Languages = 4,
     }
 
+    public enum CombatPointCategory
```
zunath, line 19: Let's put this enum in its own CombatPointCategory.cs file in the SkillService folder. The only time I double up is when the enum is directly related to the attribute. A CombatPointCategory is a property of the attribute and could have value elsewhere so it should live on its own.

SWLOR.Game.Server/Service/SkillService/SkillType.cs
```diff
@@ -20,7 +20,8 @@ public enum SkillType
             true,
             "Ability to use one-handed weapons like vibroblades, finesse vibroblades, and lightsabers.",
             true,
-            false)]
+            false,
+            CombatPointCategory.Weapons)]
```
zunath, line 24: Super nit picky, but try to use the singular to be consistent. Swap this from Weapons to Weapon please.

SWLOR.Game.Server/Service/SkillService/SkillCategoryType.cs
zunath: Also, please suffix the enum with "Type" so it's CombatPointCategoryType. The code base is slightly inconsistent with this - mostly in the Core folder - but it helps a lot with differentiating between enums and attributes.

SWLOR.Game.Server/Service/SkillService/SkillType.cs
```diff
@@ -315,6 +321,8 @@ public class SkillAttribute : Attribute
         public bool IsShownInCraftMenu { get; set; }
         public CharacterType CharacterTypeRestriction { get; set; }
 
+        public CombatPointCategory XPType { get; set; }
```
zunath, line 324: With the change to include the suffix of "Type" on CombatPointCategoryType let's go ahead and rename XPType to CombatPointCategory or something similar to make it a little clearer. XPType could be confusing - one might think it refers to combat XP, RP XP, etc.

SWLOR.Game.Server/Service/CombatPoint.cs
```diff
@@ -109,37 +109,56 @@ static void DistributeSkillXP()
...
-
-                    // Base amount of XP is determined by the player's highest-leveled skill rank versus the creature's level.
-                    var delta = npcLevel - highestRank;
-                    var baseXP = Skill.GetDeltaXP(delta);
-                    var totalPoints = (float)cpList.Sum(s => s.Value);
                     var areaBonus = GetLocalInt(GetArea(npc), "AREA_XP_BONUS_PERCENTAGE") * 0.01f;
 
-                    // Each skill used during combat receives a percentage of the baseXP amount depending on the number of combat points earned.
-                    foreach (var (skillType, cp) in cpList)
+                    foreach (CombatPointCategory cpCategory in Enum.GetValues(typeof(CombatPointCategory)))
```
zunath, line 114: I'm a LITTLE concerned about performance here. We probably need to run the profiler on it because maybe it's not an issue. But I think with some smarter caching at boot time we may be able to eliminate the 3+ iterations over the enums. I'll do some profiling on this prior to merging this in and report back.

zunath (changes requested):

SWLOR.Game.Server/Service/CombatPoint.cs
zunath: Doesn't look like this is that big of a hit. I'm fine to leave it for now. XP gains look like they're working! Nice job!

SWLOR.Game.Server/Service/CombatPoint.cs
```diff
@@ -270,16 +289,32 @@ public static void ClearRecentEnemyLevel(uint player)
...
+        }
+
         /// <summary>
         /// Adds a combat point for a player to all NPCs s/he is currently tagged on.
         /// </summary>
         /// <param name="player">The player to receiving the point.</param>
         /// <param name="skill">The skill to associate with the point.</param>
         /// <param name="amount">The number of points to add.</param>
-        /// <returns>True if at least one creature is tagged, false otherwise.</returns>
+        /// <returns>Number of creatures tagged by the player, -1 if the tracker has expired or is unavailable.</returns>
```
zunath, line 314: Sorry - can you remove this comment? No longer relevant because of the new method you added.

zunath (approved):

Issue #1451, opened 2022-07-27
Bugs #8


zunath:
Tracking bugs found in revamp testing. New thread created as the old one was getting cluttered.

Bug ticket 1: https://github.com/zunath/SWLOR_NWN/issues/1130
Bug ticket 2: https://github.com/zunath/SWLOR_NWN/issues/1359
Bug ticket 3: https://github.com/zunath/SWLOR_NWN/issues/1380
Bug ticket 4: https://github.com/zunath/SWLOR_NWN/issues/1387
Bug ticket 5: https://github.com/zunath/SWLOR_NWN/issues/1396
Bug ticket 6: https://github.com/zunath/SWLOR_NWN/issues/1402
Bug ticket 7: https://github.com/zunath/SWLOR_NWN/issues/1421

zunath:
Bug Report
The cloud effect from Bombs (Fire and Gas) do not damage the player who used them, allowing you to stand in clouds of fire and toxic gas. Epic as it is, might not be the best ;)

Bug Report
Fire/Gas Bombs from Devices do not do damage to anyone.

Bug Report
Incendiary bomb rank 1 does not seem to be doing any damage tested it on diffrent enermys

zunath:
Bug Report
Sometimes in combat, using a heal/buff ability after targetting hostiles, it instead attacks the intended target of the buff/heal ability.

Bug Report
Using a med kit while in combat makes you attack who you use the medkit on.

zunath:
Bug Report
Repeat quests of a specific type will show up as "Accepted" in the Hunters' Guild, despite not accepting them.

Bug Report
The agricultural guild is showing "Accepted" for quest I had completed prior to the last server reset, though it allows me to accept them again.

Bug Report
Possible this has already been reported. Tasks from guild masters do not update their status properly. For example if you have accepted and completed a Hunter task to take out X number of a creature it will still show ACCEPTED if it populates again and you don't have it in your quest log.

Bug Report
Some hunters quests are showing as accepted before even taking the quest, so on clicking into the quest as if to turn in the quest, it asks if you want to accept the quest, does not change to available by backing out.

zunath:
Bug Report
[...]

PR #1459, opened 2022-07-28
Fix targeted abilities switching attack targets

JohnWildkins:
Currently if you use an ability while attacking, your attack target will be switched to the target of the ability.
This is problematic in cases where you use an ability like Med Kit on a friendly while in combat, as your attack target will be switched to the friendly. Now, the attack will be set to your last attack target before using the ability instead.


JohnWildkins:
Addresses https://github.com/zunath/SWLOR_NWN/issues/1451#issuecomment-1196165778

zunath (approved):

zunath:
Oops. Good fix, thanks.

PR #1461, opened 2022-07-29
Fix Gas/Fire bomb damage, Kolto device stacking, and Throw Saber

JohnWildkins:
Fixes https://github.com/zunath/SWLOR_NWN/issues/1451#issuecomment-1196165738, https://github.com/zunath/SWLOR_NWN/issues/1451#issuecomment-1197505016

- Gas and Incendiary Bombs now properly deal damage; yes, this includes to the user!
- Kolto Grenades and Bombs now will refresh Regeneration applied to a target, rather than stacking a new regen effect.
- Throw Lightsaber now no longer fails to give you XP if you one-shot enemies with it.

### DIFF
--- a/SWLOR.Game.Server/Feature/AbilityDefinition/Devices/GasBombAbilityDefinition.cs
+++ b/SWLOR.Game.Server/Feature/AbilityDefinition/Devices/GasBombAbilityDefinition.cs
@@ -16,16 +16,9 @@ namespace SWLOR.Game.Server.Feature.AbilityDefinition.Devices
 
         private static void ApplyEffect(uint creature, int dmg)
         {
-            var activator = GetAreaOfEffectCreator(OBJECT_SELF);
-            var attackerStat = 0;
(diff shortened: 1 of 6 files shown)


SWLOR.Game.Server/Feature/AbilityDefinition/Devices/GasBombAbilityDefinition.cs
```diff
@@ -16,9 +16,15 @@ public class GasBombAbilityDefinition : ExplosiveBaseAbilityDefinition
 
         private static void ApplyEffect(uint creature, int dmg)
         {
+            var activator = GetAreaOfEffectCreator(OBJECT_SELF);
```
zunath, line 19: We may need to store these values as local variables. If the effect creator logs out of the server we'll go back to zero for the values.

I guess another option is to simply kill off the AOE when they log out? Not sure.

zunath (approved):

PR #1462, opened 2022-07-29
Fix Power Attack not applying accuracy or displaying damage

JohnWildkins:
Fixes https://github.com/zunath/SWLOR_NWN/issues/1451#issuecomment-1196166376

Power Attack damage was properly being added to base weapon damage, but not displayed anywhere, nor applying a to-hit malus. Damage is now displayed on the Character Sheet when enabled, and to-hit accuracy is appropriately reduced (and should also be displayed).

### DIFF
--- a/SWLOR.Game.Server/Feature/PerkDefinition/TwoHandedPerkDefinition.cs
+++ b/SWLOR.Game.Server/Feature/PerkDefinition/TwoHandedPerkDefinition.cs
@@ -55,13 +55,13 @@ namespace SWLOR.Game.Server.Feature.PerkDefinition
                 .Name("Power Attack")
 
                 .AddPerkLevel()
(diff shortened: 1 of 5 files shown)


SWLOR.Game.Server/Feature/PerkDefinition/TwoHandedPerkDefinition.cs
```diff
@@ -55,13 +55,13 @@ private void PowerAttack()
                 .Name("Power Attack")
 
                 .AddPerkLevel()
-                .Description("Grants the Power Attack feat which grants a +3 DMG bonus at the cost of -5 to attack roll. [Cross Skill]")
+                .Description("Grants the Power Attack feat which grants a +3 DMG bonus at the cost of -5 to accuracy. [Cross Skill]")
```
zunath, line 58: @zunath to update the spreadsheet to reflect this change

zunath (approved):

PR #1499, opened 2022-08-11
Creature management

givemedeath:
v1.0


SWLOR.Game.Server/Entity/Creature.cs
```diff
@@ -0,0 +1,24 @@
+﻿namespace SWLOR.Game.Server.Entity
+{
+    public class Creature : EntityBase
```
zunath, line 3: Let's call this a DMCreature or something to that effect. Creature is pretty ambiguous

givemedeath: Done

zunath (approved):

PR #1503, opened 2022-08-15
Emote window

givemedeath:


zunath:
@givemedeath - I did some refactoring to this PR to reduce the code duplication between chat commands and the emote window. As it was, we would have to add a new command to two places rather than just one. It now reads from the same list of commands.

Also I wanted to mention that you had a GuiList<int> of animation Ids. The only time you need to use a GuiList is if it will be bound to the UI. In all other cases you can use a regular List<>. Just a tidbit for future reference.

zunath (approved):

PR #1510, opened 2022-08-22
Area manager

givemedeath:
To use, as a DM, simply hop to one of the areas with a name starting with [Template] and spawn in some objects.
Use /am or /areas chat command to open the UI for viewing stored objects, deleting them, or resetting the entire area.

This functionality only works for areas that are "Templates". However, every module area has a corresponding template area (not including any dynamically created areas).

### DIFF
--- a/SWLOR.Game.Server/Entity/AreaTemplateObject.cs
+++ b/SWLOR.Game.Server/Entity/AreaTemplateObject.cs
@@ -1,6 +1,6 @@
 ﻿namespace SWLOR.Game.Server.Entity
 {
-    public class AreaTemplate : EntityBase
+    public class AreaTemplateObject : EntityBase
     {
         [Indexed]
         public string AreaResref { get; set; }
@@ -11,13 +11,13 @@
         public float LocationY { get; set; }
         public float LocationZ { get; set; }
         public float LocationOrientation { get; set; }
-        public AreaTemplate(string areaResRef, string name, string data, string locationX, string locationY, string locationZ, string locationOrientation)
+        public AreaTemplateObject(string areaResRef, string name, string data, string locationX, string locationY, string locationZ, string locationOrientation)
         {
             AreaResref = areaResRef;
             ObjectName = name;
             ObjectData = data;
         }
-        public AreaTemplate()
+        public AreaTemplateObject()
         {
             AreaResref = string.Empty;
             ObjectName = string.Empty;
--- a/SWLOR.Game.Server/Feature/GuiDefinition/AreaManager.cs
+++ b/SWLOR.Game.Server/Feature/GuiDefinition/AreaManager.cs
@@ -56,6 +56,17 @@ namespace SWLOR.Game.Server.Feature.GuiDefinition
                                     .BindRowCount(model => model.AreaNames);
                             });
 
+                            colAreas.AddRow(row =>
+                            {
+                                row.AddSpacer();
+
+                                row.AddButton()
+                                    .SetText("Resave All Objects")
+                                    .SetHeight(35f)
+                                    .BindOnClicked(model => model.OnClickResaveAllObjects());
+
+                                row.AddSpacer();
+                            });
                         });
 
                         row.AddColumn(colAreas =>
--- a/SWLOR.Game.Server/Feature/GuiDefinition/AreaManager.cs
+++ b/SWLOR.Game.Server/Feature/GuiDefinition/AreaManager.cs
@@ -126,7 +126,7 @@ namespace SWLOR.Game.Server.Feature.GuiDefinition
                                     .BindValue(model => model.SelectedAreaObjectName)
                                     .BindIsEnabled(model => model.IsSelectedObjectPlaceableOrCreature)
                                     .SetPlaceholder("Object Name")
-                                    .SetMaxLength(32);
+                                    .SetMaxLength(100);
                             });
 
                             colEditor.AddRow(rowEditor =>
--- a/SWLOR.Game.Server/Feature/GuiDefinition/AreaManager.cs
+++ b/SWLOR.Game.Server/Feature/GuiDefinition/AreaManager.cs
@@ -63,6 +63,7 @@ namespace SWLOR.Game.Server.Feature.GuiDefinition
                                 row.AddButton()
                                     .SetText("Resave All Objects")
                                     .SetHeight(35f)
+                                    .BindIsEnabled(model => model.IsResaveAllEnabled)
                                     .BindOnClicked(model => model.OnClickResaveAllObjects());
 
                                 row.AddSpacer();
@@ -98,6 +99,7 @@ namespace SWLOR.Game.Server.Feature.GuiDefinition
                                 row.AddButton()
                                     .SetText("Delete Object")
                                     .SetHeight(35f)
+                                    .BindIsEnabled(model => model.IsDeleteObjectEnabled)
                                     .BindOnClicked(model => model.OnClickDeleteObject());
 
                                 row.AddSpacer();
@@ -105,6 +107,7 @@ namespace SWLOR.Game.Server.Feature.GuiDefinition
                                 row.AddButton()
                                     .SetText("Reset Area")
                                     .SetHeight(35f)
+                                    .BindIsEnabled(model => model.IsResetAreaEnabled)
                                     .BindOnClicked(model => model.OnClickResetArea());
                             });
                         });
--- a/SWLOR.Game.Server/Feature/GuiDefinition/AreaManager.cs
+++ b/SWLOR.Game.Server/Feature/GuiDefinition/AreaManager.cs
@@ -34,11 +34,6 @@ namespace SWLOR.Game.Server.Feature.GuiDefinition
                             .SetText("Search")
                             .SetHeight(35f)
                             .BindOnClicked(model => model.OnClickSearch());
-
-                        row.AddButton()
-                            .SetText("TEST")
-                            .SetHeight(35f)
-                            .BindOnClicked(model => model.OnClickTest());
                     });
 
                     col.AddRow(row =>
@@ -69,11 +64,11 @@ namespace SWLOR.Game.Server.Feature.GuiDefinition
                             colAreas.SetWidth(325f);
                             colAreas.AddRow(row =>
                             {
-                                row.SetHeight(300f);
+                                row.SetHeight(450f);
 
                                 row.AddList(template =>
                                 {
-                                    template.SetHeight(300f);
+                                    template.SetHeight(450f);
 
                                     template.AddCell(cell =>
                                     {
@@ -85,15 +80,6 @@ namespace SWLOR.Game.Server.Feature.GuiDefinition
                                     });
                                 })
                                 .BindRowCount(model => model.AreaObjectList);
-                                
-                                /*
-                                row.AddTextEdit()
-                                    .SetIsMultiline(true)
-                                    .BindValue(model => model.AreaObjectList)
-                                    .SetHeight(450f)
-                                    .SetWidth(300f)
-                                    .SetIsEnabled(false);
-                                */
                             });
 
                             colAreas.AddRow(row =>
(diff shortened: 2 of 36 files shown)


SWLOR.Game.Server/Entity/AreaTemplateObject.cs
```diff
@@ -0,0 +1,28 @@
...
+        [Indexed]
+        public string AreaResref { get; set; }
+        [Indexed]
+        public string ObjectName { get; set; }
+        public string ObjectData { get; set; }
+        public float LocationX { get; set; }
+        public float LocationY { get; set; }
+        public float LocationZ { get; set; }
+        public float LocationOrientation { get; set; }
+        public AreaTemplateObject(string areaResRef, string name, string data, string locationX, string locationY, string locationZ, string locationOrientation)
```
zunath, line 14: This constructor isn't used anywhere and should be removed.

SWLOR.Game.Server/Feature/GuiDefinition/AreaManager.cs
```diff
@@ -0,0 +1,361 @@
...
+                            });
+                        });
+
+                        row.AddColumn(colObjects =>
+                        {
+                            colObjects.SetHeight(450f);
+                            colObjects.SetWidth(325f);
+                            colObjects.AddRow(row =>
+                            {
+                                //row.SetHeight(450f);
```
zunath, line 82: If this is intentionally commented out, let's go ahead and delete it to avoid confusion.

SWLOR.Game.Server/Feature/GuiDefinition/AreaManager.cs
```diff
@@ -0,0 +1,361 @@
...
+                                {
+                                    template.SetHeight(450f);
+
+                                    template.AddCell(cell =>
+                                    {
+                                        cell.AddToggleButton()
+                                            .BindText(model => model.AreaObjectList)
+                                            .BindIsToggled(model => model.AreaObjectToggled)
+                                            .BindOnClicked(model => model.OnSelectAreaObject());
+                                            //.SetWidth(300f);
```
zunath, line 94: Same thing here. Delete if intended to be commented out.

SWLOR.Game.Server/Feature/GuiDefinition/AreaManager.cs
```diff
@@ -0,0 +1,361 @@
...
+
+                            colEditor.AddRow(rowEditor =>
+                            {
+                                colEditor.AddRow(row =>
+                                {
+                                    row.SetHeight(200f);
+
+                                    row.AddList(template =>
+                                    {
+                                        //template.SetHeight(190f);
```
zunath, line 301: Same thing here. Delete if intended to be commented out.

SWLOR.Game.Server/Feature/GuiDefinition/AreaManager.cs
```diff
@@ -0,0 +1,361 @@
...
+    public class AreaManagerDefinition : IGuiWindowDefinition
+    {
+        private readonly GuiWindowBuilder<AreaManagerViewModel> _builder = new();
+
+        public GuiConstructedWindow BuildWindow()
+        {
+            _builder.CreateWindow(GuiWindowType.AreaManager)
+                .SetIsResizable(true)
+                .SetIsCollapsible(true)
+                .SetInitialGeometry(0, 0, 975f, 650f)
```
zunath, line 16: May need to reduce the default size of this window. On smaller screens it'll be difficult/impossible to select the resize section of the window. Bit of a quirk with NUI. It's safer to pick a smaller default and let the player do the resizing, especially when their UI is scaled.

SWLOR.Game.Server/Feature/GuiDefinition/ViewModel/AreaManagerViewModel.cs
```diff
@@ -0,0 +1,829 @@
...
+using SWLOR.Game.Server.Service.DBService;
+
+namespace SWLOR.Game.Server.Feature.GuiDefinition.ViewModel
+{
+    public class AreaManagerViewModel: GuiViewModelBase<AreaManagerViewModel, GuiPayloadBase>
+    {
+        private readonly List<uint> _areas = new();
+        private readonly List<uint> _objects = new();
+        private readonly List<int> _objectAppearanceIds = new();
+        private readonly List<string> _objectAppearanceNames = new();
```
zunath, line 20: This is updated but never actually used. Recommend to delete it unless this wasn't intentional.

SWLOR.Game.Server/Feature/GuiDefinition/ViewModel/AreaManagerViewModel.cs
```diff
@@ -0,0 +1,829 @@
...
+
+        private static Dictionary<int, Dictionary<int, string>> _placeableAppearanceByPage { get; } = new();
```
zunath, line 22: I don't think this needs to be a property. I'd get rid of the {get;} and treat it like a regular private field unless there's a reason for it.

SWLOR.Game.Server/Feature/GuiDefinition/ViewModel/AreaManagerViewModel.cs
```diff
@@ -0,0 +1,829 @@
...
+        private static Dictionary<int, Dictionary<int, string>> _creatureAppearanceByPage { get; } = new();
```
zunath, line 23: Same thing here - probably needs to be a field, not a property.

SWLOR.Game.Server/Feature/GuiDefinition/ViewModel/AreaManagerViewModel.cs
```diff
@@ -0,0 +1,829 @@
...
+        private static int _totalCountAppearancePlaceables;
+        private static int _totalCountAppearanceCreatures;
+
+        private bool _skipPaginationSearch;
```
zunath, line 27: This isn't used so you can delete it.

SWLOR.Game.Server/Service/AreaTemplate.cs
```diff
@@ -0,0 +1,284 @@
...
+    {
+        private static Dictionary<string, uint> TemplateAreasByResref { get; } = new();
+        private static Dictionary<uint, List<uint>> TemplateAreaCustomObjectsByArea { get; } = new();
+        private static Dictionary<int, Dictionary<int, string>> PlaceableAppearanceByPage { get; } = new();
+        private static Dictionary<int, Dictionary<int, string>> CreatureAppearanceByPage { get; } = new();
+
+        public static int TotalCountAppearancePlaceables { get; set; }
+        public static int TotalCountAppearanceCreatures { get; set; }
+
+        public const int PAGESIZE = 50;
```
zunath, line 23: Just call this PageSize. All caps constants aren't a thing with the .NET standard.

SWLOR.Game.Server/Service/AreaTemplate.cs
```diff
@@ -0,0 +1,284 @@
...
+namespace SWLOR.Game.Server.Service
+{
+    public class AreaTemplate
+    {
+        private static Dictionary<string, uint> TemplateAreasByResref { get; } = new();
+        private static Dictionary<uint, List<uint>> TemplateAreaCustomObjectsByArea { get; } = new();
+        private static Dictionary<int, Dictionary<int, string>> PlaceableAppearanceByPage { get; } = new();
+        private static Dictionary<int, Dictionary<int, string>> CreatureAppearanceByPage { get; } = new();
+
+        public static int TotalCountAppearancePlaceables { get; set; }
```
zunath, line 20: Both of these should be private as they aren't referenced outside of this service.

SWLOR.Game.Server/Service/AreaTemplate.cs
```diff
@@ -0,0 +1,284 @@
...
+            TotalCountAppearanceCreatures = actualRowCount;
+            Console.WriteLine($"Loaded {CreatureAppearanceByPage.Count} creature appearance pages ({AreaTemplate.PAGESIZE} per page)  into cache.");
+        }
+
+        [NWNEventHandler("mod_exit")]
+        public static void OnExit()
+        {
+            //return;
+
+            // For some reason I can't determine, the below is not actually working.
```
zunath, line 91: Is this comment still accurate? Is this an outstanding issue?

zunath:
image

I was able to cause an exception here. I spawned a Bird placeable which seems to have a missing appearance or something, preventing it from appearing in the area. Then I attempted to delete the object and it raised a DB exception:
```

neverwinternights-swlor-server-1  | [02:22:46 INF] C# Script 'mod_nui_event' threw an exception. Details:
neverwinternights-swlor-server-1  |
neverwinternights-swlor-server-1  | Exception type: StackExchange.Redis.RedisServerException
neverwinternights-swlor-server-1  | Message       : Syntax error at offset 54 near Id
neverwinternights-swlor-server-1  | Stacktrace:
neverwinternights-swlor-server-1  |    at StackExchange.Redis.ConnectionMultiplexer.ExecuteSyncImpl[T](Message message, ResultProcessor`1 processor, ServerEndPoint server, T defaultValue) in /_/src/StackExchange.Redis/ConnectionMultiplexer.cs:line 1873
neverwinternights-swlor-server-1  |    at NRediSearch.Client.Search(Query q) in /_/src/NRediSearch/Client.cs:line 414
neverwinternights-swlor-server-1  |    at SWLOR.Game.Server.Service.DB.Search[T](DBQuery`1 query)+MoveNext() in C:\Projects\SWLOR_NWN\SWLOR.Game.Server\Service\DB.cs:line 286
neverwinternights-swlor-server-1  |    at System.Collections.Generic.List`1..ctor(IEnumerable`1 collection)
neverwinternights-swlor-server-1  |    at System.Linq.Enumerable.ToList[TSource](IEnumerable`1 source)
neverwinternights-swlor-server-1  |    at SWLOR.Game.Server.Feature.GuiDefinition.ViewModel.AreaManagerViewModel.<OnClickDeleteObject>b__73_0() in C:\Projects\SWLOR_NWN\SWLOR.Game.Server\Feature\GuiDefinition\ViewModel\AreaManagerViewModel.cs:line 235
neverwinternights-swlor-server-1  |    at SWLOR.Game.Server.Service.Gui.HandleNuiEvents() in C:\Projects\SWLOR_NWN\SWLOR.Game.Server\Service\Gui.cs:line 236
neverwinternights-swlor-server-1  |    at SWLOR.Game.Server.Core.Internal.RunScripts(String script) in C:\Projects\SWLOR_NWN\SWLOR.Game.Server\Core\Internal.cs:line 143
neverwinternights-swlor-server-1  |
neverwinternights-swlor-server-1  |
```

SWLOR.Game.Server/Service/AreaTemplate.cs
zunath: I just tested this and reproduced the issue. The problem is the PC object is already gone from the server by the time this event is called. You can retrieve certain information about the PC but not the actual PC itself. More than likely there's a bunch of stuff that relies on the PC in the GUI framework. So you can't rely on that event firing reliably upon a disconnect.

My recommendation would be to simply remove the effect from the object directly if possible. Alternatively you can do it upon login.

zunath:
Something is wrong with the transition logic which needs to be addressed. 

1.) Teleport to the Racin' Jims template area
2.) Exit through the entrance door
3.) Character is sent to Tatooine - Mos Esper - Dowager Queen (non-template area)
4.) Using the transition in the Mos Esper area sends the player back to the template area.

zunath (changes requested):
Requesting changes for the aforementioned issues.

zunath:
I've had some time to think over this and discuss it with Ziya. We've got so many areas in the chooser as it is because of player created areas. I don't want to further overload the list so I'd like to revisit a different implementation for this feature which doesn't require us to duplicate all of the areas in the module.

Additionally, this adds a substantial amount of time to the boot up process. That time really adds up as we're debugging so it's another reason to not duplicate the areas and instead require some sort of manual process to create a copy, edit it, then save it.

We've got a thread on discord where we've begun discussing what this might look like. If you're still around/have interest feel free to join us there. For the time being I'm going to close out this PR since it's been open for a few months.

PR #1511, opened 2022-08-22
Misc. Features

givemedeath:
Implement: Add /resetwindows command #1410


givemedeath:
Going to stop adding to this PR now :)

SWLOR.Game.Server/Feature/AIDefinition/GenericAIDefinition.cs
```diff
@@ -32,485 +34,220 @@ static float CalculateAverageHP(uint creature)
...
-                return (FeatType.ElectricFist3, self);
-            }
-            if (CheckIfCanUseFeat(self, self, FeatType.ElectricFist2))
-            {
-                return (FeatType.ElectricFist2, self);
-            }
-            if (CheckIfCanUseFeat(self, self, FeatType.ElectricFist1))
-            {
-                return (FeatType.ElectricFist1, self);
+            List<FeatType> featTypes = Enum.GetValues(typeof(FeatType))
```
zunath, line 38: There's a hierarchy to the AI decisions being made here which aren't preserved by making a list out of it. I.E the AI is intended to heal itself prior to doing attacks, for example. I'm just getting into this PR but I wanted to mention it as I think this changes their behavior. 

Additionally, this AI process occurs very often - I believe on heartbeat - so the iteration over every single feat type is going to be too expensive. If this type of approach is to be used, we need to whitelist the ability types to look for rather than checking every single one of them on the NPC.

SWLOR.Game.Server/Feature/AIDefinition/GenericAIDefinition.cs
```diff
@@ -32,485 +34,220 @@ static float CalculateAverageHP(uint creature)
...
-            }
-            if (CheckIfCanUseFeat(self, target, FeatType.SpinningWhirl2))
-            {
-                return (FeatType.SpinningWhirl2, target);
-            }
-            if (CheckIfCanUseFeat(self, target, FeatType.SpinningWhirl1))
-            {
-                return (FeatType.SpinningWhirl1, target);
-            }
+        private const string CURRENT_WP_VARNAME = "WP_CUR";
```
zunath, line 139: Constants don't need to be capitalized. Can call this CurrentWaypointVariableName or something similar. Even if it's long, I prefer the names to be descriptive and to avoid acronyms in most cases. At a bare minimum remove the all caps please.

SWLOR.Game.Server/Service/Gui.cs
```diff
@@ -86,6 +86,29 @@ private static void LoadRefreshableViewModels()
...
+
+            var playerId = GetObjectUUID(player);
+            var dbPlayer = DB.Get<Player>(playerId);
+
+            foreach (var (type, _window) in _windowTemplates)
+            {
+                var window = GetPlayerWindow(player, type);
+                var resizable = JsonObjectGet(_window.Window, "resizable");
+
+                if (JsonGetInt(resizable) == 1)
```
zunath, line 102: Do we care if the window is resizable or not? Couldn't we simply reset it to the initial geometry no matter what?

SWLOR.Game.Server/Service/Spawn.cs
```diff
@@ -473,6 +473,51 @@ private static void ProcessDespawnAreas()
...
 
+        /// <summary>
+        /// On each module heartbeat, iterate over the list of areas and trigger
+        /// respawn on expired Spawns.
+        /// </summary>
+        private static void ProcessRespawnAreas()
+        {
+            var now = DateTime.UtcNow;
+
+            foreach(var _area in _activeSpawnsByArea)
```
zunath, line 484: This could be thousands of iterations every six seconds. This is going to be too heavy. We need to come up with another approach.

zunath (changes requested):
Requesting aforementioned changes

zunath:
Closing this out as it's been several months without responses to the above comments.

PR #1587, opened 2022-09-12
Fix creature aggro, holonet warnings

JohnWildkins:
- Enemies no longer consider allied corpses as allies for the purposes of sharing aggro
- Enemy corpses no longer gain enmity after dying

Those two changes above fix the current bug where corpses left on the ground will cause enemies to aggro from across the map, seemingly at random.

- Chat macros for chat commands no longer trigger Holonet warnings / put the Holonet on cooldown

### DIFF
--- a/SWLOR.Game.Server/Service/Loot.cs
+++ b/SWLOR.Game.Server/Service/Loot.cs
@@ -254,6 +254,7 @@ namespace SWLOR.Game.Server.Service
         {
             var self = OBJECT_SELF;
             SetIsDestroyable(false);
+            RemoveEffectByTag(self, "AGGRO_AOE"); // Dead creatures don't need aggro!
 
             var area = GetArea(self);
             var position = GetPosition(self);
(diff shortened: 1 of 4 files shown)


SWLOR.Game.Server/Service/Loot.cs
```diff
@@ -254,6 +254,7 @@ public static void ProcessCorpse()
         {
             var self = OBJECT_SELF;
             SetIsDestroyable(false);
+            RemoveEffectByTag(self, "AGGRO_AOE"); // Dead creatures don't need aggro!
```
zunath, line 257: This should be done in the AI service. Loot processing is a really random place to remove an AOE

JohnWildkins: to be fair, it is under `ProcessCorpse()`
to be balanced, I'll move it to crea_death_aft

zunath (approved):

PR #1619, opened 2022-12-02
Feature/base recursive language translator framework

StealthyKraken:
This feature implements the BaseRecursiveLanguageTranslator class to support explicit translation definitions for letters, segments, words and phrases, both language specific and general, all conveniently stored in 2 CSV-files. The files also support an easy definition of whether the segment has to be at the start or end of a word, sentence, etc. to be translated as such.

This is to offer a more immersive and in-universe representation of the languages seen in Star Wars and to lower the entry barrier for anyone wishing to create such definitions in the creative process of designing an unknown language from scratch. Additionally, the general definitions allow for names and such to be added to a pool of segments, that will always be translated to themselves, unless the language has its own definition for the word, i.e. jedi generally being translated into jedi, but mandoa has the word jetii for it.

Currently, Mandoa is already using the new framework with a file of about 2000 definitions. Here, the translation algorithm has been benchmarked for 300 letter input, which will be translated in a time of 2-3 ms, the result being cached for any successive listeners to the same input, the reading of which will then take ~1ms. Furthermore, Bothese, Catharese, Cheunh, Dosh and Huttese already implement the new Framework, but their definitions do not exceed what was already there before, except the usage of the general definitions.

In addition to this, the option has been added to always show the RP language in the combat log, in parallel to whichever the listener gets translated. I have also added some more commands to the /lang line of commands to make their usage more convenient.


StealthyKraken:
I have noticed a bug. Will commit a fix shortly.
grafik

StealthyKraken:
It has been fixed. 

grafik

SWLOR.Game.Server/Service/Communication.cs
```diff
@@ -401,34 +419,56 @@ public static void ProcessChatMessage()
...
                 foreach (var component in chatComponents)
                 {
+                    string languageText = "";
                     var text = component.Text;
+                    var textForCombatLog = component.Text;
 
                     if (component.IsTranslatable && language != SkillType.Basic)
                     {
-                        text = Language.TranslateSnippetForListener(sender, obj, language, component.Text);
+                        bool maxRank = false;
```
zunath, line 445: No need to declare this here. Just do it on the next line (446)

out maxRank -> out var maxRank

SWLOR.Game.Server/Service/Communication.cs
```diff
@@ -461,6 +501,10 @@ public static void ProcessChatMessage()
                 sender = originalSender;
 
                 ChatPlugin.SendMessage(finalChannel, finalMessageColoured, sender, obj);
+                var msgFCL = langMessageForCombatLog.ToString();
```
zunath, line 504: Try to avoid acronyms - use something like messageForCombatLog or similar. It's more typing but it helps with readability.

zunath (changes requested):
First off - apologies for the delay in reviewing this. I know you've been waiting for me to get to it, so I appreciate your patience.

Let me start off by saying that it's clear you put a lot of effort and work into this. I tested it out and it does seem to do what it says on the box. I started on commenting some smaller issues before I got part of the way through and had some bigger concerns I wanted to address with you before finishing the rest of it.

Here are my concerns:

1.) The translation CSV files work differently from how everything else is set up. We currently use a code-based configuration which ensures all of the data is actually in the code base. I realize this is not a recommended pattern for most apps, but it actually makes a lot of sense in NWN because of the tight coupling between the hak assets, the module, and the code. 

Having to maintain a bunch of CSV files which operate differently from the builder pattern we've already established will cause confusion for contributors. If we proceed forward with some version of this feature, I would ask that you take a look at the existing builder pattern and follow the same way for this language feature. A possible example might be something like this:
```

_builder.CreateTranslation(SkillType.Droidspeak)
    .StartCondition(...)
    .EndCondition(...)
    .SelfTranslate(...);
```

2.) I really think the vision for this feature needs to be revisited. To me, having two messages sent to players (one in the chat log and another in the combat log) is redundant, even if the setting can be toggled. I've spoken to a few different people to get their opinion on the feature and the common theme is that they wouldn't plan to use it. Obviously this wasn't a scientific poll but I'm seriously questioning whether this is worth the added overhead to process each message.

Instead, what I think might be better is if it were directly integrated into the chat message in some way. As I understand it, though, you need to have max rank in a language skill in order to get the message. If I'm wrong on this point, feel free to correct me. But if there's a way to partially translate certain things in the message based on skill level (0-20) I might be open to that.

3.) I like the shortened arguments for the /language command. If nothing else, I'd like to include this because I feel it'll help players switch languages more quickly.

4.) The last point I wanted to touch on is the complexity. I realize what you're doing is, in itself, a difficult problem to solve. But I gotta be honest, the code in your BaseRecursiveLanguageTranslator class is very complicated. I suspect some of this might be due to speed improvements you made along the way. As it sits, someone coming in fresh (i.e me) is going to have a very hard time dissecting the way this thing works. 

As it sits, I'm not sure the technical complexity and maintenance overhead are worth putting into place for a feature that may see little use. For the time being I'm going to let this PR sit to give you time to read over my thoughts. If you'd like to rework this a bit, we can discuss in Discord on what that might look like.

zunath:
Closing this out because it's been over a month without any response on the above comments.

PR #1630, opened 2022-12-11
AI update

Valo56:
All devices, force, and relevant leadership abilities now available to AI, provided they actually have the feat of course. Buffs are applied on self whenever available. Healing is applied as soon as possible on damaged targets. Because I'm an asshole NPC's can also attempt to resurrect each other (or, I suppose, try to resurrect players if they're allied).


SWLOR.Game.Server/Feature/AIDefinition/GenericAIDefinition.cs
```diff
@@ -32,7 +32,106 @@ static float CalculateAverageHP(uint creature)
             var lowestHPAllyRace = GetRacialType(lowestHPAlly);
             var allyCount = allies.Count;
             var activeConcentration = Ability.GetActiveConcentration(self).Feat;
-            
+
+            // Benevolence
+            if (CheckIfCanUseFeat(self, target, FeatType.Benevolence1, () => allyHPPercentage < 100))
```
zunath, line 37: Is it overkill to target if there's even 1 HP lost?

SWLOR.Game.Server/Feature/AIDefinition/GenericAIDefinition.cs
```diff
@@ -32,7 +32,106 @@ static float CalculateAverageHP(uint creature)
...
+            {
+                return (FeatType.Benevolence1, lowestHPAlly);
+            }
+            if (CheckIfCanUseFeat(self, target, FeatType.Benevolence2, () => allyHPPercentage < 100))
```
zunath, line 41: Benevolence 1 is going to take priority over 2 and 3. I don't think 2 or 3 will run at all if 1 is ready. This might be a place where you want to check the percentages. If they're down to 80% or higher, use 1. If they're between 50-79%, use 2, etc. Just a thought - maybe this was intentional for a reason that isn't clear to me.

SWLOR.Game.Server/Feature/AIDefinition/GenericAIDefinition.cs
```diff
@@ -32,7 +32,106 @@ static float CalculateAverageHP(uint creature)
...
+            {
+                return (FeatType.Benevolence2, lowestHPAlly);
+            }
+            if (CheckIfCanUseFeat(self, target, FeatType.Benevolence3, () => allyHPPercentage < 100))
+            {
+                return (FeatType.Benevolence3, lowestHPAlly);
+            }
+
+            // Force Heal
+            if (CheckIfCanUseFeat(self, target, FeatType.ForceHeal1, () => allyHPPercentage <= 100 && activeConcentration == FeatType.Invalid))
```
zunath, line 51: Same comment here for Force Heal, Medkit, and the remaining healing abilities.

SWLOR.Game.Server/Feature/AIDefinition/GenericAIDefinition.cs
```diff
@@ -72,6 +171,402 @@ static float CalculateAverageHP(uint creature)
                 return (FeatType.ForceStun1, target);
             }
 
+            // Adhesive Grenade
+            if (CheckIfCanUseFeat(self, target, FeatType.AdhesiveGrenade1))
```
zunath, line 175: I think at this point, you're going to want to organize these from highest to lowest. These share a recast timer so you're guaranteed to always get the lowest powered ability.

SWLOR.Game.Server/Feature/AIDefinition/GenericAIDefinition.cs
```diff
@@ -72,6 +171,402 @@ static float CalculateAverageHP(uint creature)
...
+            }
+
+            // Frenzied Shout
+            if (CheckIfCanUseFeat(self, self, FeatType.FrenziedShout))
+            {
+                return (FeatType.FrenziedShout, self);
+            }
+
+            // Soldier's Speed
+            if (CheckIfCanUseFeat(self, self, FeatType.SoldiersSpeed))
```
zunath, line 565: I'm pretty sure the AI is going to enable these auras initially and then as soon as the recast timer is up, they'll use it again which will disable it. You probably need to add a condition to ensure this doesn't happen.

zunath (changes requested):
Few things for you to look at on this. Also, got a merge conflict on the .mod file from other merges that occurred. Please resolve that as well.

SWLOR.Game.Server/Feature/AIDefinition/GenericAIDefinition.cs
Valo56: Based on my testing, there's such a long delay between when it starts taking damage and when it would use healing that this the best way to ensure they keep themselves moderately healed.

SWLOR.Game.Server/Feature/AIDefinition/GenericAIDefinition.cs
Valo56: I only give NPC's one version of these abilities, and they have sufficient FP to not worry about resource management, so it's not an issue. I could reverse them and set up different priorities but with how long it takes certain skills to actually get used I don't see a ton of value in this as by the time an NPC starts to heal after taking damage it's probably at 70% or less.

SWLOR.Game.Server/Feature/AIDefinition/GenericAIDefinition.cs
Valo56: All the NPC's I've made have one version of it. I can go through and reverse them though if needed if down the line we want NPC's to have the ability to use different tiers. However, given how useless any skill other than the max tier of that skill is (force body and med kit excluded) I don't see the value in this.

Other than consistency with other perks in aidefinition, of course, which is fair

SWLOR.Game.Server/Feature/AIDefinition/GenericAIDefinition.cs
Valo56: To be honest I haven't the faintest clue how to do that :(

SWLOR.Game.Server/Feature/AIDefinition/GenericAIDefinition.cs
zunath: I'm thinking more long term here. The same AI will be used for player companions (animals, droids, whatever) so from that perspective it'd be good to reorganize this if you don't mind. It'll also protect us if someone doesn't know about this and happens to add multiple levels. Less confusion across the board if we just get it right here in one spot.

SWLOR.Game.Server/Feature/AIDefinition/GenericAIDefinition.cs
zunath: This is going to require some refactoring. The Auras are handled from a base class which won't be accessible by the AI service. Once that's moved out, I think what we'd do here is at the start of the AI cycle, determine how many auras are active and how many they support (using existing rules). Once we have that in place it'd probably be something like:

if(CheckIfCanUseFeat(self, self, FeatType.SoldiersSpeed, () => aurasActive < maxAuras && !activeAuras.Contains(FeatType.SoldiersSpeed))

(total pseudocode here)

Because this is pretty involved I would recommend leaving out the auras from here for the time being.

SWLOR.Game.Server/Feature/AIDefinition/GenericAIDefinition.cs
Valo56: True, I'll get all these swapped around so they're in reverse order.

zunath (approved):

PR #1649, opened 2023-01-02
Attack and Force Attack crafting

Valo56:
Added Armor and Weapn Attack/Force Attack Enhancements. Accuracy enhancements included for future use, does not currently function. Recipes added (armor accuracy recipes not added, so players don't waste resources). Added new a few new NPC's for events.


SWLOR.CLI/EnhancementItemBuilder.cs
```diff
@@ -79,7 +79,9 @@ internal class EnhancementItemBuilder
             { "Shield Recharge Rate", 49 },
             { "Thermal Damage", 51 },
             { "Thermal Defense", 55 },
-            { "Module Bonus", 45 }
```
zunath, line 82: This didn't need to be removed. 45 is still Module Bonus unless I'm missing something?

Valo56: This was almost certainly in error and I can undo this change later tonight when I get home.

The only changes I recall doing here were with accuracy, attack, force attack and agility. I would have had no reason to adjust module bonuses.

zunath: I got it already, no worries. Thanks

PR #1697, opened 2023-02-12
Agriculture: Cooked Sardine

Valo56:
Issue 1683: Fix for Cooked Sardine quest. There was a recipe for Sardine Ball. This item does exist in the module but has no item properties. Cooked Sardine does exist in the module and has the appropriate item properties and resref for the indicated quest. This should fix the issue and make it possible to complete the quest.


SWLOR.Game.Server/Service/CraftService/RecipeType.cs
```diff
@@ -815,7 +815,7 @@ public enum RecipeType
         CookedBibikibo = 2091,
         SlicedSardine = 2092,
         FishBroth = 2093,
-        SardineBall = 2094,
```
zunath, line 818: Need a migration for this change. Anyone who already crafted the Sardine Ball will break after updating this.

zunath (changes requested):
Requesting changes

Valo56:
Wookies can now be much taller. Slightly taller Trandoshans, Droids and Zabraks.
Facelift to Viscara and Mon Cala station exteriors.
New Viscaran starport (uses old dockhand in same area as before, with transition to a new starport interior which may serve as a nice RP hub, no migration necessary).
Reverted changes to Sardine Ball recipe, it's back in, but now has item properties and is a Force Defense/HP alternative to the Cooked Sardine which is Physical Defense/HP. No migration necessary.
DM's will be able to spawn in all new station placeable appearances.
Two new prefabs. A last adjustment to a space prefab as its skybox was obnoxious.

Valo56:
Closing PR. Will polish and submit each change individually in a more clean manner on their own branches.

PR #1708, opened 2023-04-01
Revanite Base Revamp

ZarquonZ:
Revanite base revamp. The first pass adds in Project Praxeum, with an update anticipated in the future from Valo adding in space access. The Revanite base now includes a small mini-dungeon full of corrupted mynocks for aspirants to kill and prove themselves, as well as many new NPCs and dialogues that go into the Sith, the Force, lightsaber forms and the Revanite Order.


Module/utc/revmynock.utc.json
```diff
@@ -0,0 +1,583 @@
...
+        "Value": {
+          "type": "cexostring",
+          "value": "CZ220_LOOT_MYNOCK,100,1"
+        }
+      },
+      {
+        "__struct_id": 0,
+        "Name": {
+          "type": "cexostring",
+          "value": "QUEST_NPC_GROUP_ID"
```
zunath, line 524: You should probably assign a new NPC group to these otherwise they'll fulfill any quests assigned to the regular CZ-220 Mynocks.

SWLOR.Game.Server/Feature/SpawnDefinition/ViscaraSpawnDefinition.cs
```diff
@@ -224,5 +225,14 @@ private void VelesSewers()
                 .ReturnsHome();
         }
 
+        private void TwilightPraxeum()
+        {
+            _builder.Create("REV_MAZE")
```
zunath, line 230: Please follow the established naming scheme for spawns. Look at other spawns in this file to get an idea. It should lead with the planet, followed by a short descriptor of where it's used. 

I.E: VISCARA_REVANITE_MAZE  would suffice.

zunath (changes requested):
Requesting changes

Module/utc/revmynock.utc.json
ZarquonZ: Don't think that variable is even needed/meant for that unit type, so, just removed it

SWLOR.Game.Server/Feature/SpawnDefinition/ViscaraSpawnDefinition.cs
ZarquonZ: Changed.

zunath (approved):

PR #1710, opened 2023-04-08
Ability Scaling

Valo56:
Overall intention: To greatly increase the damage scaling of device and force abilities, to reduce the necessity of weapons to be effective in combat. Ideally it should open up tech specialist and consular/sorcerer style builds to be fully viable.

Damages are roughly scaled to be comparable to the expected damage output of a character using basic attacks while also accounting for damage losses due to misses. Abilities doing damage over time or having other utility like crowd control or debuffs will do less damage than pure damage abilities.

Somewhat mitigated the value of Force Body by introducing the Force Restore buff. At max Willpower this can allow a force user to have very consistently high access to force points, but at the cost of vitality. This means one can't just instantly recover from its use, they will suffer for having used it, and the loss of 4 points of VIT for a full minute will mean significantly more damage being taken.

Force Mind is also wrapped up in it, and both Body/Mind are now neutral powers that can both be taken by the same character. However, they are on the same cooldown, meaning you can only use one or the other. Force Mind restores Stamina at the cost of Willpower.

Force Lightning and Force Spark are unique in that they don't have an explicit force point cost. Instead they are set up to take a few force points every time they're used, but if a player doesn't have enough force points to pay that cost, it will instead begin taking health from them to cast.

Force Lightning and Force Burst are now area of effect powers with a significantly longer cooldown than Force Spark and Disturbance, which are set up to be spammable abilities. Lightning/Burst do significant damage, where Lightning damages a few targets in a very large area while Burst damages all targets in a slightly smaller area.

Eventually, when/if proper VFX are implemented to allow it it's my intent to implement Force Burst as a cylinder AoE so it's closer to Satele using it in that one SWTOR trailer. But that's a project for another day.

After this, it's my intent to go through and overhaul weapon abilities, before working on entirely new abilities.


SWLOR.Game.Server/Feature/AbilityDefinition/Devices/IncendiaryBombAbilityDefinition.cs
```diff
@@ -18,6 +18,7 @@ private static void ApplyEffect(uint creature, int dmg)
             var attackerStat = GetLocalInt(OBJECT_SELF, "DEVICE_ACC");
             var attack = GetLocalInt(OBJECT_SELF, "DEVICE_ATK");
             dmg += GetLocalInt(OBJECT_SELF, "DEVICE_DMG");
+            dmg += attackerStat / 2;
```
zunath, line 21: Instead of doing this on every AOE, you should do it one time in ExplosiveAOEImpact inside ExplosiveBaseAbilityDefinition. This will let you change it in one place instead of every file.

SWLOR.Game.Server/Feature/AbilityDefinition/Force/DisturbanceAbilityDefinition.cs
```diff
@@ -25,15 +26,51 @@ public class DisturbanceAbilityDefinition : IAbilityListDefinition
             return _builder.Build();
         }
 
-        private void Impact(uint activator, uint target, int dmg, int accDecrease, int tier, string effectTag, int dc)
+        private static void ImpactAction(uint activator, uint target, int level, Location targetLocation)
         {
+            var willBonus = GetAbilityScore(activator, AbilityType.Willpower);
+            var dmg = 0;
+            string effectTag = "";
+            int dc = 0;
```
zunath, line 34: Always prefer "var" instead of specific types. Please replace these to var for consistency.

SWLOR.Game.Server/Feature/AbilityDefinition/Force/DisturbanceAbilityDefinition.cs
```diff
@@ -25,15 +26,51 @@ public class DisturbanceAbilityDefinition : IAbilityListDefinition
             return _builder.Build();
         }
 
-        private void Impact(uint activator, uint target, int dmg, int accDecrease, int tier, string effectTag, int dc)
+        private static void ImpactAction(uint activator, uint target, int level, Location targetLocation)
         {
+            var willBonus = GetAbilityScore(activator, AbilityType.Willpower);
+            var dmg = 0;
+            string effectTag = "";
```
zunath, line 33: I'm confused about the change to the signature here. The effect tag was passed in to avoid the need to do a case statement. You're also using the wrong effect tag below. Please restructure this back to how it was before for consistency purposes and also use the tag constants that are already defined in this file.

SWLOR.Game.Server/Feature/AbilityDefinition/Force/DisturbanceAbilityDefinition.cs
```diff
@@ -25,15 +26,51 @@ public class DisturbanceAbilityDefinition : IAbilityListDefinition
...
+            int dc = 0;
+            int accDecrease = 0;
+
+            switch (level)
+            {
+                case 1:
+                    dmg = willBonus;
+                    effectTag = "Tier1Tag";
+                    dc = 8;
```
zunath, line 42: Specifically on the restructuring:

- Add the DMG value to the method signature
- Use the constant name, not a string literal for "Tier1Tag". 
- Add the DC value to the method signature.
- Add the accuracy decrease value to the method signature.

SWLOR.Game.Server/Feature/AbilityDefinition/Force/ForceBurstAbilityDefinition.cs
```diff
@@ -26,53 +27,73 @@ public class ForceBurstAbilityDefinition : IAbilityListDefinition
...
+                    
+                    DelayCommand(delay + 0.2f, () =>
+                    {
+                        ApplyEffectToObject(DurationType.Instant, EffectVisualEffect(VisualEffect.Vfx_Imp_Pulse_Wind), target);
+                    });
+
+                    CombatPoint.AddCombatPoint(activator, creature, SkillType.Force, 3);
+                    Enmity.ModifyEnmity(activator, creature, 250 * level + damage);
+                }
+                creature = GetNextObjectInShape(Shape.Sphere, RadiusSize.Medium, GetLocation(target), true , ObjectType.Creature);
```
zunath, line 95: Placing this here causes an infinite loop if the distance and reaction type condition is not met.

zunath: Actually, double check me on that. Hard to tell with GitHub formatting. You might be ok if this is outside the if statement

SWLOR.Game.Server/Feature/AbilityDefinition/Force/ForceSparkAbilityDefinition.cs
```diff
@@ -24,16 +25,52 @@ public class ForceSparkAbilityDefinition : IAbilityListDefinition
 
             return _builder.Build();
         }
-        private void Impact(uint activator, uint target, int dmg, int evaDecrease, int tier, string effectTag, int dc)
+
+        private static void ImpactAction(uint activator, uint target, int level, Location targetLocation)
```
zunath, line 29: Same comments about Disturbance apply here.

SWLOR.Game.Server/Feature/PerkDefinition/DevicesPerkDefinition.cs
```diff
@@ -93,23 +93,23 @@ private void ConcussionGrenade()
...
+                .Description("Deals electrical DMG equal to your Perception Score to all creatures within range of explosion. Consumes explosives on use.")
                 .RequirementSkill(SkillType.Devices, 5)
                 .RequirementCharacterType(CharacterType.Standard)
                 .Price(2)
                 .DroidAISlots(2)
                 .GrantsFeat(FeatType.ConcussionGrenade1)
 
                 .AddPerkLevel()
-                .Description("Deals 10 electrical DMG to all creatures within range of explosion. Also has an 8DC reflex check to inflict Knockdown for 3 seconds. Consumes explosives on use.")
+                .Description("Deals 15 electrical DMG, scaling with your Percepiton Score, to all creatures within range of explosion. Also has an 8DC reflex check to inflict Knockdown for 3 seconds. Consumes explosives on use.")
```
zunath, line 104: Typo: Percepiton -> Perception

SWLOR.Game.Server/Feature/PerkDefinition/DevicesPerkDefinition.cs
```diff
@@ -64,22 +64,22 @@ private void FragGrenade()
...
-                .Description("Deals 6 fire DMG to all creatures within range of explosion. Consumes explosives on use.")
+                .Description("Deals fire DMG equal to your Perception Score to all creatures within range of explosion. Consumes explosives on use.")
                 .Price(2)
                 .DroidAISlots(2)
                 .RequirementCharacterType(CharacterType.Standard)
                 .GrantsFeat(FeatType.FragGrenade1)
 
                 .AddPerkLevel()
-                .Description("Deals 10 fire DMG to all creatures within range of explosion. Also has an 8DC reflex check to inflict Bleeding. Consumes explosives on use.")
+                .Description("Deals 20 fire DMG, scaling with your Percepiton Score, to all creatures within range of explosion. Also has an 8DC reflex check to inflict Bleeding. Consumes explosives on use.")
```
zunath, line 74: Typo: Percepiton -> Perception

SWLOR.Game.Server/Feature/PerkDefinition/DevicesPerkDefinition.cs
```diff
@@ -93,23 +93,23 @@ private void ConcussionGrenade()
...
+                .Description("Deals 15 electrical DMG, scaling with your Percepiton Score, to all creatures within range of explosion. Also has an 8DC reflex check to inflict Knockdown for 3 seconds. Consumes explosives on use.")
                 .Price(3)
                 .DroidAISlots(3)
                 .RequirementSkill(SkillType.Devices, 30)
                 .RequirementCharacterType(CharacterType.Standard)
                 .GrantsFeat(FeatType.ConcussionGrenade2)
 
                 .AddPerkLevel()
-                .Description("Deals 16 electrical DMG to all creatures within range of explosion. Also has a 12DC reflex check to inflict Knockdown for 3 seconds. Consumes explosives on use.")
+                .Description("Deals 30 electrical DMG, scaling with your Percepiton Score, to all creatures within range of explosion. Also has a 12DC reflex check to inflict Knockdown for 3 seconds. Consumes explosives on use.")
```
zunath, line 112: Typo: Percepiton -> Perception

SWLOR.Game.Server/Feature/PerkDefinition/DevicesPerkDefinition.cs
```diff
@@ -153,23 +153,23 @@ private void IonGrenade()
                 .Name("Ion Grenade")
 
                 .AddPerkLevel()
-                .Description("Deals 4 electrical DMG to all enemies within range of explosion. Consumes explosives on use.")
+                .Description("Deals electrical DMG equal to your Percepiton Score to all enemies within range of explosion. Deals bonus damage to droids. Consumes explosives on use.")
```
zunath, line 156: Typo: Percepiton -> Perception

SWLOR.Game.Server/Feature/PerkDefinition/DevicesPerkDefinition.cs
```diff
@@ -153,23 +153,23 @@ private void IonGrenade()
...
                 .RequirementSkill(SkillType.Devices, 5)
                 .RequirementCharacterType(CharacterType.Standard)
                 .Price(2)
                 .DroidAISlots(2)
                 .GrantsFeat(FeatType.IonGrenade1)
 
                 .AddPerkLevel()
-                .Description("Deals 8 electrical DMG to all enemies within range of explosion. Also has a 10DC fortitude check to inflict stun to droids for 6 seconds. Consumes explosives on use.")
+                .Description("Deals 8 electrical DMG, scaling with your Percepiton Score, to all enemies within range of explosion. Deals bonus damage to droids. Also has a 10DC fortitude check to inflict stun to droids for 6 seconds. Consumes explosives on use.")
```
zunath, line 164: Typo: Percepiton -> Perception

SWLOR.Game.Server/Feature/PerkDefinition/DevicesPerkDefinition.cs
```diff
@@ -153,23 +153,23 @@ private void IonGrenade()
...
                 .Price(3)
                 .DroidAISlots(3)
                 .RequirementSkill(SkillType.Devices, 20)
                 .RequirementCharacterType(CharacterType.Standard)
                 .GrantsFeat(FeatType.IonGrenade2)
 
                 .AddPerkLevel()
-                .Description("Deals 14 electrical DMG to all enemies within range of explosion. Also has a 14DC fortitude check to inflict stun to droids for 6 seconds. Consumes explosives on use.")
+                .Description("Deals 14 electrical DMG, scaling with your Percepiton Score, to all enemies within range of explosion. Deals bonus damage to droids. Also has a 14DC fortitude check to inflict stun to droids for 6 seconds. Consumes explosives on use.")
```
zunath, line 172: Typo: Percepiton -> Perception

SWLOR.Game.Server/Feature/PerkDefinition/DevicesPerkDefinition.cs
```diff
@@ -297,21 +297,21 @@ private void IncendiaryBomb()
                 .Name("Incendiary Bomb")
 
                 .AddPerkLevel()
-                .Description("Creates a fire field at the explosion site, dealing 4 fire DMG to all creatures who enter the area of effect for 20 seconds. Consumes explosives on use.")
+                .Description("Creates a fire field at the explosion site, dealing 4 fire DMG, scaling with your Percepiton Score, to all creatures who enter the area of effect for 20 seconds. Consumes explosives on use.")
```
zunath, line 300: There's a bunch more typos below. Copy/pasting can be bad ;) Please fix them all

SWLOR.Game.Server/Feature/PerkDefinition/ForcePerkDefinition.cs
```diff
@@ -344,23 +344,21 @@ private void ForceBurst()
         }
         private void ForceMind()
         {
-            _builder.Create(PerkCategoryType.ForceLight, PerkType.ForceMind)
+            _builder.Create(PerkCategoryType.ForceUniversal, PerkType.ForceMind)
                 .Name("Force Mind")
 
                 .AddPerkLevel()
-                .Description("Converts 25% of the user's FP into HP.")
+                .Description("Grants Stamina Point regeneration based on your Willpower Score, at the cost of your Willpower.")
```
zunath, line 351: I like the addition of the scaling info but we should probably be consistent with naming. "STM" is used elsewhere so I would replace "Stamina Point" with "STM" or just "Stamina" since we don't ever make a reference to points to my knowledge.

SWLOR.Game.Server/Feature/PerkDefinition/ForcePerkDefinition.cs
```diff
@@ -416,31 +414,31 @@ private void ForceLightning()
                 .Name("Force Lightning")
 
                 .AddPerkLevel()
-                .Description("Deals 12 DMG to a single target.")
+                .Description("Deals DMG equal to your Willpower Score to up to 5 targets in a radius. Consumes Force Points, but if none remain, will consume Health instead.")
```
zunath, line 417: For consistency let's replace Health with HP

SWLOR.Game.Server/Feature/PerkDefinition/ForcePerkDefinition.cs
```diff
@@ -450,23 +448,21 @@ private void ForceLightning()
 
         private void ForceBody()
         {
-            _builder.Create(PerkCategoryType.ForceDark, PerkType.ForceBody)
+            _builder.Create(PerkCategoryType.ForceUniversal, PerkType.ForceBody)
                 .Name("Force Body")
 
                 .AddPerkLevel()
-                .Description("Converts 25% of the user's HP into FP.")
+                .Description("Grants Force Point regeneration based on your Willpower Score, at the cost of your Vitality.")
```
zunath, line 455: Same comment here about Force Points. Calling it FP will make it consistent across the board.

SWLOR.Game.Server/Feature/StatusEffectDefinition/ForceRestoreStatusEffectDefinition.cs
```diff
@@ -0,0 +1,106 @@
...
+            ForceBody2(builder);
+            ForceMind1(builder);
+            ForceMind2(builder);
+
+            return builder.Build();
+        }
+
+        private void ForceRestore(uint source, uint target, bool forceBody, int level)
+        {
+            int willBonus = GetAbilityScore(source, AbilityType.Willpower);
```
zunath, line 27: Prefer "var" to specific  types.

SWLOR.Game.Server/Feature/StatusEffectDefinition/ForceRestoreStatusEffectDefinition.cs
```diff
@@ -0,0 +1,106 @@
...
+
+            int forceAmount = willBonus * level / 4;
+            int staminaAmount = (2 * level) + willBonus * level / 4;
+
+            if (forceBody == true)
```
zunath, line 32: The "== true" is unnecessary, please remove.

SWLOR.Game.Server/Feature/UsePerkFeat.cs
```diff
@@ -239,8 +239,11 @@ void CompleteActivation(string id, float abilityRecastDelay)
                 Activity.ClearBusy(activator);
 
                 // Moved during casting or activator died. Cancel the activation.
-                if (GetLocalInt(activator, id) == (int)ActivationStatus.Interrupted || GetCurrentHitPoints(activator) <= 0)
+                if (GetLocalInt(activator, id) == (int)ActivationStatus.Interrupted || GetCurrentHitPoints(activator) <= 0 || GetCurrentAction(activator) != ActionType.Invalid)
+                {
+                    Activity.ClearBusy(activator);
```
zunath, line 244: Busy status has already been cleared on line 239, this isn't necessary.

SWLOR.Game.Server/Feature/UsePerkFeat.cs
```diff
@@ -239,8 +239,11 @@ void CompleteActivation(string id, float abilityRecastDelay)
                 Activity.ClearBusy(activator);
 
                 // Moved during casting or activator died. Cancel the activation.
-                if (GetLocalInt(activator, id) == (int)ActivationStatus.Interrupted || GetCurrentHitPoints(activator) <= 0)
+                if (GetLocalInt(activator, id) == (int)ActivationStatus.Interrupted || GetCurrentHitPoints(activator) <= 0 || GetCurrentAction(activator) != ActionType.Invalid)
```
zunath, line 242: This is going to interrupt casting if any action is being performed, including an automatic attack by the player if they happen to be in melee. Have you confirmed that queued abilities (i.e weapon abilities) are not impacted by this change?

zunath (changes requested):
Requesting changes

SWLOR.Game.Server/Feature/UsePerkFeat.cs
Valo56: I was able to queue pistol abilities in testing without any difficulty.

SWLOR.Game.Server/Feature/AbilityDefinition/Force/ForceBurstAbilityDefinition.cs
Valo56: It is outside the if statment, inside the while statement.
image

SWLOR.Game.Server/Feature/AbilityDefinition/Force/ForceSparkAbilityDefinition.cs
```diff
@@ -1,5 +1,5 @@
 ﻿using System.Collections.Generic;
-using SWLOR.Game.Server.Core;
+using System.Reflection.Emit;
```
zunath, line 2: This shouldn't be needed. Please remove

Valo56: Done.

zunath (approved):

PR #1711, opened 2023-04-09
Increased Max Height for certain races

Valo56:
Wookie, Droid, Trandoshan, Zabrak and Kel Dor PC's can now be taller.


SWLOR.Game.Server/Feature/GuiDefinition/ViewModel/AppearanceEditorViewModel.cs
```diff
@@ -920,7 +920,18 @@ private void LoadItemParts()
         {
             var scale = GetObjectVisualTransform(_target, ObjectVisualTransform.Scale);
             const float Increment = 0.01f;
-            const float MaximumScale = 1.15f;
+            var MaximumScale = 1.15f;
```
zunath, line 923: This implementation is too much of a hack. We have racial appearance classes for each of the races. I.E: KelDorRacialAppearance.cs

A better implementation would be to add MaximumScale and MinimumScale to those classes (and the interface) and use those values. That way, all values are in a single place.

Lucky for you I'm already doing all this for you. ;)

zunath:
Handling the fixes in separate PR due to lack of permissions. https://github.com/zunath/SWLOR_NWN/pull/1715

PR #1720, opened 2023-04-17
Custom Area Transitions

Valo56:
Added /destination and /settag commands for DM's to allow for ease of setting up transitions. Adjusted teleport script to also check for object tags.


SWLOR.Game.Server/Feature/ChatCommandDefinition/DMChatCommand.cs
```diff
@@ -406,10 +406,66 @@ private void SetLocalVariable()
 
                     SendMessageToPC(user, "Local string set: " + variableName + " = " + value);
                 });
+
+            _builder.Create("settag")
+                .Description("Sets a local tag on a target.")
+                .Permissions(AuthorizationLevel.DM, AuthorizationLevel.Admin)
+                .AvailableToAllOnTestEnvironment()
```
zunath, line 413: I don't want this exposed to all players on the test server. SetTag is a pretty dangerous function to give to DMs, let alone players. Definitely remove this line please.

SWLOR.Game.Server/Feature/ChatCommandDefinition/DMChatCommand.cs
```diff
@@ -406,10 +406,66 @@ private void SetLocalVariable()
...
+                    if (args.Length <= 0)
+                    {
+                        return "Missing arguments. Format should be: /SetTag <VALUE>. Example: /SetTag DUNGEON_ENTRANCE";
+                    }
+
+                    return string.Empty;
+                })
+                .Action((user, target, location, args) =>
+                {
+                    if (!GetIsObjectValid(target))
```
zunath, line 426: I'm on the fence about this one. Tags are used in some parts of the scripting and my concern is if this command is used on the wrong object we could run into systemic problems. Is there anything you can do to restrict its usage to the custom area transitions? Maybe abstract it away a bit more so that it's like:

/SetTransition MY_TRANSITION

and requiring the target object be a specific type of placeable?

SetTag is a little too exposed for my liking.

SWLOR.Game.Server/Feature/ChatCommandDefinition/DMChatCommand.cs
```diff
@@ -406,10 +406,66 @@ private void SetLocalVariable()
...
+            _builder.Create("destination")
+                .Description("Sets the DESTINATION localstring to the given value on a target.")
+                .Permissions(AuthorizationLevel.DM, AuthorizationLevel.Admin)
+                .AvailableToAllOnTestEnvironment()
+                .RequiresTarget()
+                .Validate((user, args) =>
+                {
+                    if (args.Length <= 0)
+                    {
+                        return "Missing arguments. Format should be: /destionation <VALUE>. Example: /destination EventEntrance";
```
zunath, line 447: Typo: destionation -> destination

SWLOR.Game.Server/Feature/ChatCommandDefinition/DMChatCommand.cs
```diff
@@ -406,10 +406,66 @@ private void SetLocalVariable()
...
+                        SetTag(target, args[0]);
+
+                        SendMessageToPC(user, "Tag set to: " + args[0] + ".");
+                    }
+                });
+
+            _builder.Create("destination")
+                .Description("Sets the DESTINATION localstring to the given value on a target.")
+                .Permissions(AuthorizationLevel.DM, AuthorizationLevel.Admin)
+                .AvailableToAllOnTestEnvironment()
```
zunath, line 441: This is another one I wouldn't want exposed to players on the test server.

SWLOR.Game.Server/Feature/ChatCommandDefinition/DMChatCommand.cs
```diff
@@ -406,10 +406,66 @@ private void SetLocalVariable()
...
+                        return "Missing arguments. Format should be: /destionation <VALUE>. Example: /destination EventEntrance";
+                    }
+
+                    return string.Empty;
+                })
+                .Action((user, target, location, args) =>
+                {
+                    if (!GetIsObjectValid(target))
+                    {
+                        SendMessageToPC(user, "Target is invalid. Please target a placeable.");
```
zunath, line 456: This is misleading because it will actually work on any type of object so long as it's valid. This means you could target items, players, etc. If you really intend this to be used on placeables, please change the if statement to do a check on GetObjectType instead.

SWLOR.Game.Server/Feature/ChatCommandDefinition/DMChatCommand.cs
```diff
@@ -406,10 +406,66 @@ private void SetLocalVariable()
...
+                    }
+                    else
+                    {
+                        SetEventScript(target, EventScript.Placeable_OnUsed, "teleport");
```
zunath, line 460: This is another situation where I think we need to be more restrictive for this command. You can really hose world objects if the command is misused by the DMs.

The more I'm thinking about it, perhaps the workflow should be a single command like /CreateTeleporter DESTINATION_TAG which spawns a new placeable instead of modifying an existing one. Then you can change the graphic used after it's been spawned?

SWLOR.Game.Server/Feature/ChatCommandDefinition/DMChatCommand.cs
Valo56: Oh, yeah, true. Removing.

SWLOR.Game.Server/Feature/ChatCommandDefinition/DMChatCommand.cs
Valo56: I can require it so that it fails if the resref of the placeable doesn't match the Teleport Object placeable. That should sidestep any possible issues.

SWLOR.Game.Server/Feature/ChatCommandDefinition/DMChatCommand.cs
Valo56: I've restricted its use only to the Teleport Object placeable so there should be no issues with DM's accidentally messing up an area or some critical object.

SWLOR.Game.Server/Feature/ChatCommandDefinition/DMChatCommand.cs
```diff
@@ -406,10 +406,64 @@ private void SetLocalVariable()
...
+                        SetLocalString(target, "DESTINATION", args[0]);
+                        SendMessageToPC(user, "Applied teleport script to object. Destination tag set to " + args[0] + ".");
+                    }
+                }); 
         }
 
-        private void SetPortrait()
-        {
+            private void SetPortrait()
```
zunath, line 465: Some unnecessary tabbing here - can you revert?

SWLOR.Game.Server/Feature/ChatCommandDefinition/DMChatCommand.cs
```diff
@@ -406,10 +406,64 @@ private void SetLocalVariable()
...
+                    }
+                    else
+                    {
+                        SetTag(target, args[0]);
+
+                        SendMessageToPC(user, "Tag set to: " + args[0] + ".");
+                    }
+                });
+
+            _builder.Create("destination")
```
zunath, line 437: For consistency can we call this tpdestination to signify it should be used in conjunction with tptag?

SWLOR.Game.Server/Feature/ChatCommandDefinition/DMChatCommand.cs
```diff
@@ -406,10 +406,64 @@ private void SetLocalVariable()
...
+                .Description("Sets the DESTINATION localstring to the given value on a target.")
```
zunath, line 438: Maybe make this less technical? Something like "Sets the tag of the destination on a teleport object" or similar?

SWLOR.Game.Server/Feature/ChatCommandDefinition/DMChatCommand.cs
Valo56: Yeah. Is tpdest okay? Just so it's short. We already have the teleport to waypoint as tpwp for instance.

SWLOR.Game.Server/Feature/ChatCommandDefinition/DMChatCommand.cs
Valo56: oops, yeah

SWLOR.Game.Server/Feature/ChatCommandDefinition/DMChatCommand.cs
Valo56: Yeah will reword tonight and update this

zunath (approved):

PR #1737, opened 2023-05-23
Enmity Rebalance Changes

Sytrath:
Makes Tanking a little easier, and off-tanking (though require micromanaging) possible. This eases the gameplay off of Supports a little and makes fights, dependent on the presence of a tank, more chaotic as enemies will likely spread out without one. Will need a look at over the coming days to see if it's working well or not, and if not, a re-check.

Balance feedback is requested: The goal is to try and invite change to the current system, not that my changes are automatically correct because I am oh so smart ha ha. I'd like to pass this around a little bit and gauge people's thoughts.


Valo56:
Having gone over this, I agree with the changes. I also agree with the nerf to Benevolence's overall healing. Time will tell if it requires further nerfs. There's nothing here I find problematic.

Sytrath:
Thank you very much Valo for looking at my first balance patch! I am interested in providing more balance discussions for the community in the future, to try and take the workload off of you. Let me know if you would like to have any game design discussions! :)

PR #1738, opened 2023-05-23
Veles Revamp

ZarquonZ:
Veles revamp. Most of the credit goes to Grik for his incredible builds of the maps. Honorary mention to Whyleth et other DMs who chipped in here and there. Changes include but are not limited to:
+ Veles has been changed. There's a central garden, a market, and shops for Smithery, Engineering, etc. NPCs now walk around, with Czerka patrols that can provide direction. The Cantina is back near the center of Veles and should be a more interesting place to RP at.
+ A new 'Slums' zone is added, with new interiors, and NPCs hinting at underlying tensions between Czerka, anti-Czerkan forces, and smugglers.
+ The sewers has been expanded to include a new Smugglers' base, with multiple entrances and its own keys.
+ Many interiors have been reworked; there's no longer a blanket R&D for crafting, but instead there's separate buildings and lore for Smithery, Fabrication, and Engineering. There's also a new HoloNews Net office.
+ All NPCs in Veles should now have their own descriptions and dialogues.


ZarquonZ:
Closing this to make a fresh one.

PR #1756, opened 2023-06-01
Fishing Fix

Valo56:
Fix for fishing bug resulting in players being stuck busy when attempting to fish from a fishing point that had already been exhausted.

This happens because after the fishing point is set to dead, it remains on screen for a few seconds before despawning. The same would happen with mining except those are sent to a z coordinate of -10 or something after being rendered dead, preventing players from doing anything with them as they're underground.

By having a player's busy state cleared when they do this it avoids any issues.


SWLOR.Game.Server/Service/Fishing.cs
```diff
@@ -284,6 +284,7 @@ void CheckPosition(uint player, Vector3 startPosition, string attemptId)
             if (!GetIsObjectValid(fishingPoint) || GetIsDead(fishingPoint))
             {
                 SendMessageToPC(player, "This fishing point has been exhausted.");
+                Activity.ClearBusy(player);
```
zunath, line 287: I'm skeptical of this because you're changing the state of the PC when, in reality, they shouldn't ever get to the Busy state in the first place.

I suspect the problem isn't with clearing the status but rather the GetIsDead check we're doing on line 284. I'll bet that only applies to creatures. Maybe it should be swapped from GetIsDead to GetCurrentHitPoints() < 0 instead. I'd recommend trying that before we start mucking with player state.

zunath (changes requested):
Requesting changes/open discussions above

zunath:
Closing this out due to inactivity. Please feel free to reopen once comments above are addressed.

PR #1757, opened 2023-06-01
Provoke 2 Fix

Valo56:
Provoke 2 will no longer attempt to provoke your own beast.


SWLOR.Game.Server/Feature/AbilityDefinition/Armor/ProvokeAbilityDefinition.cs
```diff
@@ -80,7 +80,7 @@ private void Provoke2()
                         if (GetDistanceBetweenLocations(GetLocation(nearest), location) > 8f)
                             break;
 
-                        if (!GetIsPC(nearest))
+                        if (!GetIsPC(nearest) && BeastMastery.IsPlayerBeast(activator) == false)
```
zunath, line 83: Do a henchman check here instead. This won't work for droids as written. GetAssociate is the function you need.

zunath: Also, I think this should be caught in the enmity service rather than here. Otherwise every place we adjust enmity could run the risk of including henchmen.

zunath (changes requested):
Requesting changes/open discussions above

zunath:
Closing this out due to inactivity. Please feel free to reopen once comments above are addressed.

PR #1770, opened 2023-06-27
Dantooine upate

Scorchys12:
*Added another player buildable area near the jungles *Added some more areas *updated areas with some content
Tamable beasts need to be set up


Module/nss/dangym1.nss
```diff
@@ -0,0 +1,12 @@
+void main()
```
zunath, line 1: No NWScripts please.

zunath (changes requested):
Requesting changes

PR #1771, opened 2023-06-29
Updated veles with:  slums district, Industial sector, commercial / h…

Scorchys12:
…igh end area. shops updated with a small new area. Two more  prefabs thanks to King.


zunath (changes requested):
Merge conflict on this one. Please resolve at your convenience.

PR #1776, opened 2023-07-05
*Ate the signs in veles, hooked up new areas: Slums, commerical block…

Scorchys12:
…s, Snazzy's burger shop updated.

*Added achievements to new areas.


zunath:
Please add the following to this PR:

- Add MAP_KEY_ITEM_ID, value 40 to all of the new exterior areas to ensure the Veles map will work on them. Look at Viscara - Veles variables for an example.
- Please remove all spawn points for the colonists. Their pathfinding is causing server lag because there are too many of them. Please fix this for all new areas as well as the existing Veles exterior area.

Also, I originally denied the additional areas when Ziya submitted them because they really have zero content in them. They look great, don't get me wrong, but there really is nothing to do in them besides walk around. I'm pretty on the fence about this at the moment but if you can give a good justification for keeping them in their current state I could maybe understand it. Otherwise I think it might be best we leave them inaccessible normally until there's actually something to do in them.

zunath (changes requested):
Requesting changes above

KingArthur129:
I agree with the nothing to do in these areas, I would recommend leaving them as plot areas for the time being. I would remove the 'closed' signs to these areas and just retain them as a shield 'wall' so people don't always see the closed signs.

Scorchys12:
I agree in leaving it inaccessible for now, they're nice RP areas but i don't know how to put in interesting quests at this time. Will close this for now and revisit later.

PR #1779, opened 2023-07-14
Add new method GetPerkAdjustedAbilityScore in combat.cs 

Garrbear758:
Check for the relevant damage ability score based on weapon type, perks, and stances. Apply the new method to all relevant abilities.

Abilities Affected:
-Force Leap
-Saber Strike
-Crippling Shot
-Double Shot
-Explosive Toss
-Piercing Toss
-Quick Draw
-Circle Slash
-Double Strike
(Leg Sweep and Slam now use the new method as well, but they had inline code to support flurry already, but crushing style / mastery will now also apply to abilities.)


zunath:
Made a few stylistic fixes to your PR. Had to submit separately due to permissions issues on GitHub. Closing this out as it has been merged. Thank you!

PR #1809, opened 2023-12-03
feature/outpost-hope-revamp

PrimetheGrime:
Area overhauls on Outpost Hope, Appearance changes for npcs, Addition of decolight placeables in white, orange, green, blue, cyan


Module/nss/zep_use_chair.nss
```diff
@@ -1,20 +1,2 @@
...
-
-
-void main()
-{
-    object oChair = OBJECT_SELF;
-    if (!GetIsObjectValid(GetSittingCreature(OBJECT_SELF))) {
-        AssignCommand(GetLastUsedBy(), ActionSit(oChair));
-    }
-}
+{WalkWaypoints()
```
zunath, line 1: No NWScript changes allowed. Please remove.

zunath (changes requested):
Requesting changes

Module/nss/zep_use_chair.nss
```diff
@@ -1,20 +0,0 @@
-//::///////////////////////////////////////////////
```
zunath, line 1: Hang on. You don't want to delete it because that will completely remove it from the module. In this case, it was already in the module so you want to add it back with the original code. Just copy/paste this into a new zep_use_chair.nss file and commit/push it up. Should fix the issue.

PR #1812, opened 2023-12-04
Player Self-Portrait Change Command

Scorchys12:
Allows a player to change their own portrait without need for DM assistance.


zunath:
This feature will be implemented into NUI. No reason for it to be a chat command.

PR #1815, opened 2024-01-29
DM Item Management Tools

Valo56:
Added chat commands to:
-add specific item properties
-remove all item properties
-toggle plot on and off for an item


zunath:
I have some concerns over the implementation - primarily around the fact that the command to add properties isn't dynamic. Every time we add something new, we'll have to remember to update this command. Given the maintenance overhead I'd like to shelve this approach for now in favor of a more fitting NUI to handle it more dynamically. Seeing as Valo has left the community, it'll fall on one of us to set this up. Going to close out this PR and reference it later when we begin work on its replacement.

PR #1816, opened 2024-01-29
Player Aura Fix

Valo56:
Creates a chat command performing the same function as a respawn for players with regards to removing and reapplying auras. Allows for players to reapply leadership buffs after dying and being revived, going to space, or any other issues that may emerge down the line with regard to auras.


zunath:
This is fixing a symptom of the issue, not the underlying problem. Let's fix the bug, not put a bandaid on it.

PR #1819, opened 2024-02-02
Space Update

Valo56:
For now all I've done is added chat commands for DM's to see ship stats and repair targeted ships.

More to come in time.


SWLOR.Game.Server/Feature/ChatCommandDefinition/DMChatCommand.cs
```diff
@@ -1019,5 +1020,68 @@ private void GetScale()
...
+                                $"Explosive Defense: {targetStatus.ExplosiveDefense} \n" +
+                                $"Skill Level: {level}");
+                        }
+                    }
+                });
+        }
+
+        private void RepairShip()
+        {
+            _builder.Create("repairship")
```
zunath, line 1068: Could you just bake this in to the existing /rez command?

SWLOR.Game.Server/Feature/GuiDefinition/ViewModel/ShipManagementViewModel.cs
```diff
@@ -923,6 +945,23 @@ private void LoadShip()
                     LowPower8Tooltip = string.Empty;
                 }
 
+                ConfigurationVisible = shipDetail.ConfigurationNodes >= 1;
+
+                module = ship.Status.ConfigurationModules.ContainsKey(1)
```
zunath, line 950: I would do a count check here. If we add a second or third configuration module in the future, this will break if the user puts a module in those slots instead of the first one.

SWLOR.Game.Server/Feature/GuiDefinition/ViewModel/ShipManagementViewModel.cs
```diff
@@ -1396,6 +1504,11 @@ private void ProcessLowPower(int slot)
             ProcessLowPower(8);
         };
 
+        public Action OnClickConfiguration() => () =>
+        {
+            ProcessConfiguration(8);
```
zunath, line 1509: Why is this 8?

SWLOR.Game.Server/Feature/ChatCommandDefinition/DMChatCommand.cs
Valo56: The advantage here is I figure DM's can apply damage manually to ships with this command as well, or to only repair a certain amount. At present there's no way for DM's to do either of those things with ships. I can bake it into rez instead if you'd prefer to not have another chat command.

SWLOR.Game.Server/Feature/GuiDefinition/ViewModel/ShipManagementViewModel.cs
Valo56: Copy/paste error. Will fix.

SWLOR.Game.Server/Feature/GuiDefinition/ViewModel/ShipManagementViewModel.cs
Valo56: I've implemented the code for up to 8 config slots. Maybe with corvettes when I add them I'll allow for up to two configuration slots to be used on them. It would open up a lot of customization options I think.

SWLOR.Game.Server/Feature/ChatCommandDefinition/DMChatCommand.cs
```diff
@@ -1019,5 +1020,68 @@ private void GetScale()
...
+                .Description("Restores a ship's armor, shield and capacitor by the indicated amount.")
+                .Permissions(AuthorizationLevel.DM, AuthorizationLevel.Admin)
+                .AvailableToAllOnTestEnvironment()
+                .RequiresTarget()
+                .Action((user, target, location, args) =>
+                {
+                    if (GetIsDM(target) != true && GetIsDMPossessed(target) != true)
```
zunath, line 1075: Small stylistic thing, but would prefer this to be either:

!GetIsDM(target) && !GetIsDMPossessed(target)

or

GetIsDM(target) == false && GetIsDMPossessed(target) == false

The first one is more preferred over the second, but != true is weird :D

SWLOR.Game.Server/Feature/ChatCommandDefinition/DMChatCommand.cs
zunath: I can kind of see the value of restoring by a specific amount, but if you really think that will be used we can keep it. The only thing I'd ask is that if a number isn't provided, do a full recovery on the ship. Right now it'll do nothing unless a number is entered, and I know people are going to forget.

SWLOR.Game.Server/Feature/GuiDefinition/ShipManagementDefinition.cs
```diff
@@ -310,6 +310,81 @@ public GuiConstructedWindow BuildWindow()
...
+                    {
+                        row.AddLabel()
+                            .SetText("Configuration Slot")
+                            .SetHeight(20f);
+                    });
+
+                    col.AddRow(row =>
+                    {
+                        row.AddButtonImage()
+                         .BindImageResref(model => model.Configuration1Resref)
```
zunath, line 323: Let's fix the tab formatting on this stuff. One tab for each dot method, new line in between elements.

SWLOR.Game.Server/Feature/GuiDefinition/ViewModel/ShipManagementViewModel.cs
zunath: This isn't doing what I thought it was, disregard.

SWLOR.Game.Server/Feature/LootTableDefinition/SpaceLootTableDefinition.cs
```diff
@@ -20,20 +20,17 @@ public class SpaceLootTableDefinition : ILootTableDefinition
 
         private void Tier1()
         {
-            _builder.Create("SPACE_SYBIL")
-                .AddItem("elec_ruined", 20)
-                .AddItem("ref_tilarium", 3);
+            _builder.Create("SPACE_BASIC_T1")
```
zunath, line 23: Please review these loot tables for anything that needs to be marked as a rare item. Right now none of them are, so Treasure Hunter won't work.

SWLOR.Game.Server/Feature/PerkDefinition/GeneralPerkDefinition.cs
```diff
@@ -1,9 +1,11 @@
 ﻿using System.Collections.Generic;
 using SWLOR.Game.Server.Core.NWScript.Enum;
 using SWLOR.Game.Server.Entity;
+using SWLOR.Game.Server.Enumeration;
```
zunath, line 4: These changes aren't necessary. Please remove.

SWLOR.Game.Server/Feature/RecipeDefinition/EngineeringRecipeDefinition/ModuleRecipes.cs
```diff
@@ -176,7 +176,61 @@ private void Tier1()
...
+                .Level(9)
+                .Quantity(1)
+                .RequirementPerk(PerkType.StarshipBlueprints, 1)
+                .EnhancementSlots(RecipeEnhancementType.Module, 1)
+                .Component("ref_tilarium", 5)
+                .Component("elec_ruined", 3);
+
+            // Mirrored Plating 1
+            _builder.Create(RecipeType.ThermalArmor1, SkillType.Engineering)
+				.Category(RecipeCategoryType.ShipModule)
```
zunath, line 192: Bunch of weird formatting with the tabs here.

SWLOR.Game.Server/Feature/RecipeDefinition/EngineeringRecipeDefinition/ModuleRecipes.cs
```diff
@@ -494,6 +602,60 @@ private void Tier3()
 				.Component("ref_idailia", 5)
 				.Component("elec_good", 3);
 
+			// Hypermatter Injector III
```
zunath, line 605: I need these recipes added to the Bible too.

SWLOR.Game.Server/Feature/ShipDefinition/NPCShipDefinition.cs
```diff
@@ -21,365 +21,629 @@ public class NPCShipDefinition : IShipListDefinition
 
         private void Tier1()
         {
-            _builder.Create("NPC_Sybil")
-                .ItemResref("npc_sybil")
-                .Name("NPC - Sybil")
-                .Appearance(AppearanceType.SWLORShipNeutralDropship)
```
zunath, line 27: The appearance on all of these ships is missing. Can you double check on this? Are they actually showing the right models?

Valo56: The right models display, the appearance in the code seems to be unnecessary for npc's, it's just pc's that need it.

SWLOR.Game.Server/Feature/ShipModuleDefinition/MiningLaserModuleDefinition.cs
```diff
@@ -55,10 +55,16 @@ private void MiningLaser(string itemTag, string name, string shortName, int requ
                         return "This mining laser is not powerful enough to harvest that asteroid.";
                     }
 
+                    if (GetLocalBool(target, "BEING_MINED") == true)
```
zunath, line 58: Drop the == true, not needed.

Please put "BEING_MINED" into a constant.

SWLOR.Game.Server/Feature/ShipModuleDefinition/MiningLaserModuleDefinition.cs
```diff
@@ -132,6 +138,7 @@ private void MiningLaser(string itemTag, string name, string shortName, int requ
                             var xp = Skill.GetDeltaXP(delta);
 
                             Skill.GiveSkillXP(activator, SkillType.Piloting, xp);
+                            SetLocalBool(target, "BEING_MINED", false);
```
zunath, line 141: I'd put this variable outside the PC check. If, somehow, a non-PC uses a mining laser the rock will be locked forever.

SWLOR.Game.Server/Feature/ShipModuleDefinition/MissileLauncherModuleDefinition.cs
```diff
@@ -88,10 +79,36 @@ private void PerformAttack(uint activator, uint target, int dmg, int attackBonus
                 .ValidTargetType(ObjectType.Creature)
                 .PowerType(ShipModulePowerType.High)
                 .RequirePerk(PerkType.OffensiveModules, requiredLevel)
-                .Recast(recast)
-                .Capacitor(capacitor)
+                .Recast(12f)
+                .ValidationAction((activator, activatorShipStatus, target, targetShipStatus, moduleBonus) =>
+                {
+                    var item = GetItemPossessedBy(activator, "ship_missile");
```
zunath, line 85: Another situation where I'd set "ship_missile" to a constant so you are limiting the use of magic strings.

SWLOR.Game.Server/Feature/ShipModuleDefinition/ShipArmorModuleDefinition.cs
```diff
@@ -0,0 +1,123 @@
...
+                .Name(name)
+                .ShortName(name)
+                .Texture("iit_ess8_145")
+                .Description($"Improves a ship's overall defenses by {armorBoost} at the cost of {armorBoost} evasion.")
+                .PowerType(ShipModulePowerType.Low)
+                .RequirePerk(PerkType.DefensiveModules, requiredLevel)
+                .EquippedAction((creature, shipStatus, moduleBonus) =>
+                {
+                    shipStatus.ThermalDefense += armorBoost + moduleBonus;
+                    shipStatus.EMDefense+= armorBoost + moduleBonus;
```
zunath, line 109: Really minor but can you put a space before the plus sign here?

SWLOR.Game.Server/Feature/ShipModuleDefinition/ShipConfigurationModuleDefinition.cs
```diff
@@ -0,0 +1,142 @@
...
+        }
+
+        private void ShipConfig(string itemTag, string name, string texture, int requiredLevel, int armor, int shield, int thermalAttack, int thermalDefense, int ionAttack, int ionDefense, int explosiveAttack, int explosiveDefense, int accuracy, int evasion, int industrial)
+        {
+            _builder.Create(itemTag)
+                .Name(name)
+                .ShortName(name)
+                .Texture(texture)
+                .Description($"Provides several base stats to a ship to allow it to fulfill a role and operate at full capacity: \n" +
+                $"Armor: +{armor} + Module Bonus * 4 \n" +
```
zunath, line 87: Since this is being displayed to players, I'd recommend using "x4" instead of "* 4" simply because non-programmers might not understand what that notation means.

SWLOR.Game.Server/Service/SpaceService/ShipBuilder.cs
```diff
@@ -135,6 +135,19 @@ public ShipBuilder LowPowerNodes(int lowPowerNodes)
             return this;
         }
 
+        /// <summary>
+        /// Sets the number of ship configuration nodes on this ship.
+        /// By default ships should have 1 configuration node.
+        /// </summary>
+        /// <param name="shipConfigurationNodes">The number of low power nodes to set.</param>
```
zunath, line 142: Description for this param is wrong

SWLOR.Game.Server/Feature/ShipModuleDefinition/CapitalWeaponsModuleDefinition.cs
```diff
@@ -0,0 +1,270 @@
...
+                    var attackRoll = Random.D10(1);
+                    var attackType = "";
+                    var attackBonus = activatorShipStatus.ThermalDamage;
+                    var attackerStat = GetAbilityScore(activator, AbilityType.Perception);
+                    var attack = Stat.GetAttack(activator, AbilityType.Perception, SkillType.Piloting, attackBonus);
+                    var defenseBonus = targetShipStatus.ThermalDefense * 2;
+                    var defense = Stat.GetDefense(target, CombatDamageType.Thermal, AbilityType.Vitality, defenseBonus);
+                    if (attackRoll < 7)
+                    {
+                        attackType = "laser";
```
zunath, line 65: Strings are very error prone and difficult to maintain. Please set up an enumeration - you can make it private within this class. I.E:

private enum AttackType
{
    Invalid = 0,
    Laser = 1,
...etc
}

SWLOR.Game.Server/Feature/ShipModuleDefinition/CapitalWeaponsModuleDefinition.cs
```diff
@@ -0,0 +1,270 @@
...
+                        0);
+
+                    var chanceToHit = Space.CalculateChanceToHit(activator, target);
+                    var roll = Random.D100(1);
+                    var isHit = roll <= chanceToHit;
+
+                    var sound = EffectVisualEffect(VisualEffect.Vfx_Ship_Blast);
+                    var missile = EffectVisualEffect(VisualEffect.Mirv_StarWars_Bolt2);
+
+                    if (attackType == "laser")
```
zunath, line 102: This method is getting very long. I would recommend splitting it up a bit. Perhaps a method for laser attacks, another for ion attacks, etc. There needs to be some organization here because it's going to be hard to maintain.

SWLOR.Game.Server/Feature/ShipModuleDefinition/CapitalWeaponsModuleDefinition.cs
```diff
@@ -0,0 +1,270 @@
...
+                    {
+                        if (isHit)
+                        {
+                            AssignCommand(activator, () =>
+                            {
+                                ApplyEffectToObject(DurationType.Instant, sound, target);
+                                ApplyEffectToObject(DurationType.Instant, missile, target);
+
+                                DelayCommand(0.3f, () =>
```
zunath, line 111: What's the purpose of the delay?

SWLOR.Game.Server/Feature/ShipModuleDefinition/CapitalWeaponsModuleDefinition.cs
```diff
@@ -0,0 +1,270 @@
...
+                    var defense = Stat.GetDefense(target, CombatDamageType.Thermal, AbilityType.Vitality, defenseBonus);
+                    if (attackRoll < 7)
+                    {
+                        attackType = "laser";
+                    }
+                    else if (attackRoll < 9)
+                    {
+                        attackType = "ion";
+                        dmg += dmg / 3;
+                        attackBonus = activatorShipStatus.EMDamage;
```
zunath, line 71: This assignment never gets used.

SWLOR.Game.Server/Feature/ShipModuleDefinition/CapitalWeaponsModuleDefinition.cs
```diff
@@ -0,0 +1,270 @@
...
+                        defenseBonus = targetShipStatus.EMDefense * 2;
+                        defense = Stat.GetDefense(target, CombatDamageType.EM, AbilityType.Vitality, defenseBonus);
+                    }
+                    else
+                    {
+                        attackBonus = activatorShipStatus.ExplosiveDamage;
```
zunath, line 77: Another assignment that never gets used.

SWLOR.Game.Server/Feature/ShipModuleDefinition/CapitalWeaponsModuleDefinition.cs
```diff
@@ -0,0 +1,270 @@
...
+                        if (isHit)
+                        {
+                            AssignCommand(activator, () =>
+                            {
+                                var effect = EffectBeam(VisualEffect.Vfx_Beam_Mind, activator, BodyNode.Chest);
+                                ApplyEffectToObject(DurationType.Temporary, effect, target, 1.0f);
+                                ApplyEffectToObject(DurationType.Instant, EffectVisualEffect(VisualEffect.Vfx_Ion_Shot), activator);
+                            });
+
+                            DelayCommand(0.1f, () =>
```
zunath, line 144: Another delay here which I'm not understanding. Unless you absolutely need it for some reason I would remove it.

SWLOR.Game.Server/Feature/ShipModuleDefinition/CapitalWeaponsModuleDefinition.cs
```diff
@@ -0,0 +1,270 @@
...
+                            {
+                                AssignCommand(activator, () =>
+                                {
+                                    var shieldDamage = int.Min(damage, targetShipStatus.Shield);
```
zunath, line 148: This might be fixed by removing the delay, but you should be aware of a bit of a quirk in the language. The "damage" variable here can actually be changed outside the scope of this inner AssignCommand call. This page has a better explanation of the problem: https://www.jetbrains.com/help/resharper/AccessToModifiedClosure.html

The solution is to set damage to a new variable within the AssignCommand call.

SWLOR.Game.Server/Feature/ShipModuleDefinition/CapitalWeaponsModuleDefinition.cs
```diff
@@ -0,0 +1,270 @@
...
+                        AssignCommand(activator, () =>
+                        {
+                            ApplyEffectToObject(DurationType.Instant, EffectVisualEffect(VisualEffect.Vfx_Ship_Trp), activator);
+                            ApplyEffectToObject(DurationType.Instant, EffectVisualEffect(VisualEffect.Mirv_Torpedo, !isHit), target);
+                        });
+
+                        DelayCommand(delay, () =>
+                        {
+                            ApplyEffectToObject(DurationType.Instant, EffectVisualEffect(VisualEffect.Fnf_Fireball, !isHit, 0.5f), target);
+                            if (isHit)
```
zunath, line 207: Another closure issue.

SWLOR.Game.Server/Feature/ShipModuleDefinition/CapitalWeaponsModuleDefinition.cs
```diff
@@ -0,0 +1,270 @@
...
+                        var combatLogMessage = Combat.BuildCombatLogMessage(activator, target, attackId, chanceToHit);
+                        Messaging.SendMessageNearbyToPlayers(target, combatLogMessage, 60f);
+
+                        Enmity.ModifyEnmity(activator, target, damage);
+                        CombatPoint.AddCombatPoint(activator, target, SkillType.Piloting);
+                    }
+                    else
+                    {
+                        var targetDistance = GetDistanceBetween(activator, target);
+                        var targetLocation = GetLocation(target);
```
zunath, line 194: This is unused.

SWLOR.Game.Server/Feature/ShipModuleDefinition/CapitalWeaponsModuleDefinition.cs
```diff
@@ -0,0 +1,270 @@
...
+                                attackBonus = activatorShipStatus.ThermalDamage;
+                                defenseBonus = targetShipStatus.ThermalDefense * 2;
+                                defense = Stat.GetDefense(nearbyTarget, CombatDamageType.Thermal, AbilityType.Vitality, defenseBonus);
+                                damage = Combat.CalculateDamage(attack, dmg, attackerStat, defense, defenderStat, 0) / 3;
+                                AssignCommand(activator, () =>
+                                    {
+                                        ApplyEffectToObject(DurationType.Instant, sound, target);
+                                        ApplyEffectToObject(DurationType.Instant, missile, target);
+
+                                        DelayCommand(0.3f, () =>
```
zunath, line 241: More delay command calls I don't understand.

SWLOR.Game.Server/Feature/ShipModuleDefinition/CapitalWeaponsModuleDefinition.cs
```diff
@@ -0,0 +1,270 @@
...
+                    var count = 0;
+                    var nearbyTarget = GetFirstObjectInShape(Shape.Sphere, RadiusSize.Gargantuan, GetLocation(activator), true);
+                    while (GetIsObjectValid(nearbyTarget) && count < 5)
+                    {
+                        if (GetIsReactionTypeHostile(nearbyTarget, activator))
+                        {
+                            chanceToHit = Space.CalculateChanceToHit(activator, nearbyTarget);
+                            roll = Random.D100(1);
+                            isHit = roll <= chanceToHit;
+                            if (isHit == true && nearbyTarget != target)
```
zunath, line 229: == true is unnecessary

SWLOR.Game.Server/Feature/ShipModuleDefinition/CapitalWeaponsModuleDefinition.cs
Valo56: The delay is the same as for combat lasers I believe (on phone atm hard to double check). Just has the damage and other effects occur after allowing time for the vfx to conclude.

SWLOR.Game.Server/Feature/RecipeDefinition/EngineeringRecipeDefinition/ModuleRecipes.cs
Valo56: I think I've fixed the tabs now, please let me know if not, but none of this weirdness was showing for me in VS in the first place.

SWLOR.Game.Server/Feature/RecipeDefinition/EngineeringRecipeDefinition/ModuleRecipes.cs
Valo56: I'll get a full list of updated stats and recipes your way soon. If you'd like it sooner than later I can get on that otherwise I was planning on waiting until wider testing was underway and further changes are unlikely.

Valo56:
Mostly full list of changes:
Global Cooldown: There is a 2 second wait time for both PC's and NPC's between module activations.

T1 merchant-bought ship appearances changed from foray/aurek to other small ships.

New Ships:
-Aurek Strikefighter
-Sith Fighter
-Basilisk War Droid
These require a recipe that will be available from DM's.

Cutlass Fighter
Think of it as an unaligned military fighter.

Corvettes:
Thranta (Neutral, Sith, Republic color schemes)
Corellian Gunboat (kind of looks like the firefly)
Chiss Trireme
Terminus Class Corsair (uses the Kusari Gunship appearance)
Hutt Corvette
CZC Armored Transport (big beefy boy)

Different corvette skins are purely cosmetic, there is no stat difference between them.
Each corvette has 3 high slots and 4 low slots. 750 armor and 750 shield. 350 capacitor and a 30 second shield recharge rate.
This may not seem like much, but capital ships use (and can only use) special capital ship modules that have very powerful effects.

Ship Configuration Slots:
Each ship has one ship config slot. These provide a variety of base stats to a ship, intended to bring attack and defense values more in line with values you'd see in ground combat, and lengthening the time of each engagement so it's not a game of one or two hit kills.
-Fighter: All around good, but isn't great anywhere.
-Bomber: High attack. Good defenses and health pools. Low explosive defense, but high damage. 
-Interceptor: High evasion. Poor defenses and health pools. High explosive defense, average defenses elsewhere. 
-Industrial: Offers very l ow defenses, and no offensive/evasive bonuses. Very high health pools. Adds to a ship's industrial level, improving the strength of repair modules and the yield of mining lasers.

New Subcapital Modules
Beam Cannon: Subcapital. Attack 4 times in quick succession. Repeated hits can ramp up your damage on subsequent hits.
Proton Bomb Launcher: Short range, requires a Proton Bomb (craftable). Can only target capital ships. Bypasses shields, and does even more damage if the target doesn't have shields.
-Hypermatter Injector: Uses Hypermatter Capsule consumables to quickly restore capacitor.

Subcapital Adjustments
Combat Laser: No capacitor cost. Average damage.
Ion Cannon: High damage, but only 50% damage output vs armor. On doing armor damage, inflicts penalties to move speed and agility.
Missile Launcher: No capacitor cost. Above average damage. Slowish cooldown but not as slow as before. Requires missile item to fire. Does bonus damage on unshielded targets.

New NPC Pirate Clans, each with a boss NPC and 6 different ship types beyond that - Fighter, Interceptor, Bomber, Cargo Ship, Gunship, Defense Platforms:
T1: Septra Khanate
T2: Krakenborne
T3: Snowdrift Syndicate
T4: Clan Zel Arun
T5 Red Hand (T5 Boss drops capship recipes 66% of the time)
T6: Sith Empire (7 different tiers of boss NPC's, from corvettes to the Harrower itself) (all T6 bosses drop recipes)

Module bonuses for weapons have a more significant impact on weapon performance. It doesn't just affect Attack rating but contributes to DMG as well.

PC ships of most tiers have had their number of slots and overall HP/SP buffed.

Capital Ship Modules:
-Advanced Thruster: Low slot. Reduces capacitor in exchange for significant evasion buff.
-Redundant Shield Generator: Greatly improved max SHP, but slower shield recharge.
-Reinforced Plating: Significant bonus to ship defense scores, but reduced evasion.
-Dedicated Targeting Sensor Array: Significant bonus to accuracy and attack, at the cost of similar penalty to evasion. 
-Ship Config: Only three types. Industrial Nexus, Skirmisher (good shields and regen), Warship (high defense, attack, accuracy)

Cont.
Assault Concussion Missile Launcher: Requires very expensive missiles to fire. Bypasses shields, and does huge damage. But can only target capships.
Bulwark Shield Generator: Generate massive aggro. Restore shields for self and allies. Raise defenses of allies.
Laser Cannon Battery: Fire once every 2 seconds for 18 seconds at 50% of all nearby hostile targets.
Quad Laser Cannon: Fire 4-6 times in quick succession, based on tier of Quad Laser. Good at engaging subcapital ships.
Repair Field Generator: Regenerate hull HP for self and allies over a period of time.
Storm Cannon: A beam of high energy particles strips away a target's shields, doing massive shield damage. It imposes significant penalties on damaging armor, moreso than an Ion Cannon, but deals even less hull damage.
Turbolasers: Three tiers. Each has a 50% miss chance vs subcapital ships.
Heavy Turbolaser: High damage. Single shot.
Dual Turbolaser: Two shots over two seconds. Slightly less damage per shot but more overall.
Quad Turbolaser: Four shots over fours econds. High cap cost. Very high overall damage output.
New DM Commands: /shipstats and /repairship
Displays stats for ships. Repairs ships. Self explanatory. As DM's have no way of doing either right now this should help us run events.

New interiors:
Corvettes and the new fighters each use new interior areas. Fighters are very small areas with little room to move around. Basilisks give you a miniature hangar bay. Corvettes are massive areas repurposed from the capship event area many of you are familiar with.

NPC AI:
NPC ships will use the new modules. Their default distance to remain apart from each other in space is now 15 meters instead of 5.

SWLOR.Game.Server/Feature/ShipModuleDefinition/BulwarkShieldGeneratorModuleDefinition.cs
```diff
@@ -0,0 +1,72 @@
...
+                .Texture("iit_ess_075")
+                .Description(description)
+                .PowerType(ShipModulePowerType.High)
+                .Capacitor(25)
+                .Recast(12f)
+                .CapitalClassModule()
+                .CanTargetSelf()
+                .ActivatedAction((activator, activatorShipStatus, target, targetShipStatus, moduleBonus) =>
+                {
+                    var industrialBonus = Space.GetShipStatus(activator).Industrial;
```
zunath, line 42: This has the potential to cause a null reference exception. Space.GetShipStatus can return a null in certain cases. I believe you need to add a null check here (and anywhere else) to bail out early if you don't have a valid ShipStatus object.

zunath: Actually, looking at it closer you shouldn't need to call this at all. You already have activatorShipStatus available to you. Just use that and you don't need the null check.

Valo56: Adjusted this.

SWLOR.Game.Server/Feature/ShipModuleDefinition/ShipArmorModuleDefinition.cs
```diff
@@ -0,0 +1,123 @@
...
+                {
+                    shipStatus.ThermalDefense -= 2 * armorBoost + moduleBonus;
+                });
+        }
+
+        private void IonArmor(string itemTag, string name, string shortName, int armorBoost, int requiredLevel)
+        {
+            _builder.Create(itemTag)
+                .Name(name)
+                .ShortName(name)
```
zunath, line 63: This should be shortName

SWLOR.Game.Server/Feature/ShipModuleDefinition/ShipArmorModuleDefinition.cs
```diff
@@ -0,0 +1,123 @@
...
+                {
+                    shipStatus.EMDefense -= 2 * armorBoost + moduleBonus;
+                });
+        }
+
+        private void ExplosiveArmor(string itemTag, string name, string shortName, int armorBoost, int requiredLevel)
+        {
+            _builder.Create(itemTag)
+                .Name(name)
+                .ShortName(name)
```
zunath, line 82: This should be shortName

SWLOR.Game.Server/Feature/ShipModuleDefinition/ShipArmorModuleDefinition.cs
```diff
@@ -0,0 +1,123 @@
...
+                {
+                    shipStatus.ExplosiveDefense -= 2 * armorBoost + moduleBonus;
+                });
+        }
+
+        private void HeavyArmor(string itemTag, string name, string shortName, int armorBoost, int requiredLevel)
+        {
+            _builder.Create(itemTag)
+                .Name(name)
+                .ShortName(name)
```
zunath, line 101: This should be shortName

zunath (approved):

PR #1838, opened 2024-05-11
Adrenal Stim Tweaks

Valo56:
Updated Adrenal Stim to not have reduced effects due to its own Willpower penalty. This brings it in line with Force Mind.

### DIFF
--- a/SWLOR.Game.Server/Feature/StatusEffectDefinition/AdrenalStimEffectDefinition.cs
+++ b/SWLOR.Game.Server/Feature/StatusEffectDefinition/AdrenalStimEffectDefinition.cs
@@ -1,5 +1,6 @@
 ﻿using System.Collections.Generic;
 using SWLOR.Game.Server.Core.NWScript.Enum;
+using SWLOR.Game.Server.Core.NWScript.Enum.VisualEffect;
 using SWLOR.Game.Server.Service;
 using SWLOR.Game.Server.Service.StatusEffectService;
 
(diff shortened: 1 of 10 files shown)


SWLOR.Game.Server/Feature/StatusEffectDefinition/AdrenalStimEffectDefinition.cs
```diff
@@ -20,6 +20,10 @@ public class AdrenalStimStatusEffectDefinition : IStatusEffectListDefinition
         private void AdrenalStim(uint source, uint target, int level)
         {
             var willBonus = GetAbilityModifier(AbilityType.Willpower, source);
+            if (source == target)
+            {
+                willBonus == level * 2;
```
zunath, line 25: This wasn't tested ;)

How do I know? It's a compilation error.

taelon875:
Looking over the numbers on this adrenal stim 3 will give back 9 STM a tick on self regardless of WIL. While a max 30 WIL character would give 11 to someone else. It looks rather strong for self STM regen.  Also how much STM is a standard using round on average?  With this and a SOC STM regen aura (yes that is a thing people forget about) a standard could be getting back what... 17 STM a round.)

Valo56:
Force Body/Mind restore Will Modifier * Level (which goes up to 2) / 4. 30 willpower on a forcerer thus gives 30 * 2 / 4 = 15 per round. Which is a lot and could probably do with being lowered.

Adrenal stim granting 10 or so per round doesn't seem overwhelming in comparison. Granted, all of this was based on way back when tick actions were often 30 seconds to a minute apart or more, and you'd be lucky to get more than two ticks off of a single Force Body. Now that it's consistently per round, it may be better to bring the regeneration rate closer to the 4-8 range per round.

With that said, Standards tend to use more of their resources than Force Sensitives. Standards require high willpower to have any useful healing, unless they dish out a lot of STM on higher tier first aid perks. Compare that to force users that either get by just fine with low level first aid perks due to their naturally higher willpower, and have access to other low cost healing through Heal, Drain and Benevolence.

The availability of food, better tick consistency, means the Adrenal Stim perk and its force counterparts could probably stand to be lowered. So I'll look into doing that.

And yes, definitely testing each change I make.. even if it's "okay yeah this is working I'll just change this one other value and send it in" aaaaaaaaaaa *embarrassedscream.gif*

taelon875:
Yeah, I cant really speak to STM use, i hardly use it.  I think the point of it way more effective to use it on yourself vs someone else though should be considered.  Also back in the day i am pretty sure force users could only take force body or force mind not both. Which is why the values also might have been higher.

Valo56:
I've gone through and adjusted (and tested) these changes now, all working as expected. Ranges for the Stim and Force Mind/Body will tend toward the 4-8 range. Very high Willpower characters can hit 8/round, but everyone else will tend toward 4-6. The durations have also been decreased considerably, which combined with the high cooldown each of these abilities have will mean that restoring resources in longer fights is possible but that overall food buffs and teamwork through regen auras (or pet buffs I suppose) is going to shine a lot more.

taelon875:
I would argue the nerf on force mind isnt good at all. As a pure mage character i will lose more FP then i will ever make back with current costs.  As there is no other sources of FP regen other then this and natural ticks which only happen ever 30ish seconds.

zunath (approved):

PR #1839, opened 2024-05-12
Immobilized While Casting

Valo56:
When a player activates an ability, they are immobilized for 1.5 seconds. No other effects are applied during this. The old requirement that players not have any action queued up when an ability finished was removed, as this was janky (WASD did not cause a failure, only click moving).

Most abilities have a cast time beyond 1.5 seconds, notably tame and resurrect abilities. Players being able to move during this time however is fine. The point is to reduce/eliminate the Benny Hill combat system we have now so that we can begin balancing around a system that doesn't have people running in circles.

The immobilize only applies to PC's who are not in space mode, as this isn't an issue in space.

### DIFF
--- a/SWLOR.Game.Server/Feature/UsePerkFeat.cs
+++ b/SWLOR.Game.Server/Feature/UsePerkFeat.cs
@@ -273,6 +273,11 @@ namespace SWLOR.Game.Server.Feature
             CheckForActivationInterruption(activationId, position);
             SetLocalInt(activator, activationId, (int)ActivationStatus.Started);
 
+            if (GetIsPC(activator) && !Space.IsPlayerInSpaceMode(activator))
+            {
+                ApplyEffectToObject(DurationType.Temporary, EffectCutsceneImmobilize(), activator, 1.5f);
+            }
+
             var executeImpact = ability.ActivationAction == null 
                 ? true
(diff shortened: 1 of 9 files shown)


SWLOR.Game.Server/Feature/UsePerkFeat.cs
```diff
@@ -273,6 +273,11 @@ void CompleteActivation(string id, float abilityRecastDelay)
             CheckForActivationInterruption(activationId, position);
             SetLocalInt(activator, activationId, (int)ActivationStatus.Started);
 
+            if (GetIsPC(activator) && !Space.IsPlayerInSpaceMode(activator))
+            {
+                ApplyEffectToObject(DurationType.Temporary, EffectCutsceneImmobilize(), activator, 1.5f);
```
zunath, line 278: Any real significance to 1.5 seconds? Do you want to account for heavy armor here? I.E longer immobilize time compared to light?

Valo56: For abilities with cast times 1.5 seconds is usually the lowest they go (pretty sure that's what throw lightsaber is at). 1.5 seconds should be enough time for, if someone is trying to kite an enemy who is right behind them, for them to both be stopped and for the enemy chasing them to begin a flurry of attacks.

zunath (approved):

PR #1844, opened 2024-06-02
Space update tweaks

Valo56:
Additional tweaks. Mostly based on feedback or minor documentation issues (such as item descriptions being inaccurate). Also updated fighter interiors.

Also updated the immobilize when casting feature from one of the last PR's.  There was an issue with it stuttering. With slow in place it should feel more like a more gradual deceleration to players. When the slow ends, the player's speed will be recalculated to prevent issues with dash and other move speed buffs breaking.


SWLOR.Game.Server/Feature/RecipeDefinition/EngineeringRecipeDefinition/ModuleRecipes.cs
```diff
@@ -264,7 +264,7 @@ private void Tier1()
                 .Component("ref_tilarium", 3)
                 .Component("elec_ruined", 2);
 
-            // Interceptor Config 1
+            // Basic Interceptor Config
             _builder.Create(RecipeType.InterceptorConfig1, SkillType.Engineering)
```
zunath, line 268: If you're going to rename the items, can you make sure the enumerations also match? It's difficult to change it after this releases so I'd like to make the correction now if at all possible.

SWLOR.Game.Server/Feature/UsePerkFeat.cs
```diff
@@ -275,7 +275,12 @@ void CompleteActivation(string id, float abilityRecastDelay)
 
             if (GetIsPC(activator) && !Space.IsPlayerInSpaceMode(activator))
             {
+                ApplyEffectToObject(DurationType.Temporary, EffectSlow(), activator, 1.5f);
```
zunath, line 278: What's the purpose of both a slow and an immobilize at the same time?

SWLOR.Game.Server/Feature/UsePerkFeat.cs
```diff
@@ -275,7 +275,12 @@ void CompleteActivation(string id, float abilityRecastDelay)
...
                 ApplyEffectToObject(DurationType.Temporary, EffectCutsceneImmobilize(), activator, 1.5f);
+                DelayCommand(1.5f, () =>
```
zunath, line 280: Let's make this 1.6f to ensure it happens after the other effects are intended to expire.

zunath (changes requested):
Requesting above changes

SWLOR.Game.Server/Feature/UsePerkFeat.cs
taelon875: I asked about this as well, if i remember the discussion the slow was to limit APR to 1. And the immobilize was to make it so they cant move.

Valo56: Immobilize keeps them from moving, ensuring kiting can't be done so easily. However, due to the way NWN handles paralysis, it doesn't take effect smoothly, and this has resulted in people stuttering to a stop. The slow smooths this out, making it feel less janky, with the move speed restored to the level it should be at the end. I'll adjust the timing to make sure there's no overlap between slowing and the timer expiring in the case of weird lag related issues.

Valo56: Also a slow alone isn;t sufficient as even while moving at a slow pace it's possible to disrupt an attacker's attacks per round when they're forced to catch up.

Valo56:
I've updated the above. I've not been able to work out the migration issue, however.

zunath (approved):

PR #1858, opened 2024-07-27
There What you asked for

taelon875:
Only the alser and useperkfixes

### DIFF
--- a/SWLOR.Game.Server/Feature/UsePerkFeat.cs
+++ b/SWLOR.Game.Server/Feature/UsePerkFeat.cs
@@ -223,9 +223,9 @@ namespace SWLOR.Game.Server.Feature
                     currentPosition.Z != originalPosition.Z)
                 {
                     RemoveEffectByTag(activator, "ACTIVATION_VFX");
-
                     PlayerPlugin.StopGuiTimingBar(activator, string.Empty);
                     Messaging.SendMessageNearbyToPlayers(activator, $"{GetName(activator)}'s ability has been interrupted.");
+                    SetLocalInt(activator, activationId, (int)ActivationStatus.Interrupted);
                     return;
                 }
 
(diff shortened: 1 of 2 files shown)


SWLOR.Game.Server/Feature/UsePerkFeat.cs
```diff
@@ -273,15 +273,6 @@ void CompleteActivation(string id, float abilityRecastDelay)
             CheckForActivationInterruption(activationId, position);
             SetLocalInt(activator, activationId, (int)ActivationStatus.Started);
 
-            if (GetIsPC(activator) && !Space.IsPlayerInSpaceMode(activator))
-            {
-                ApplyEffectToObject(DurationType.Temporary, EffectSlow(), activator, 1.5f);
-                DelayCommand(1.6f, () =>
-                {
-                    Stat.ApplyPlayerMovementRate(activator);
```
zunath, line 281: Are you sure this isn't needed?

taelon875: 95% sure?  The only things that are in space that would trigger this are things with activation delays... And nothing but mining lasers have those. And honestly you shouldnt be moving while mining. So i felt it solved two issues there.

zunath: Well, no. You're still immobilizing the player on line 287 below which affects movement speed. The reapplication of the movement rate is there to make sure that immobilization (or any kind of slow) doesn't permanently reduce speed. Can you double check this by turning on Dash, using an ability, then making sure your movement rate is still the same as it was before the ability usage?

taelon875: I mean 287 is the one i added after this was removed, and in my testing I didn't notice any reduction in speed. I will apply a movement speed back after.

taelon875:
Okay... After a few hours I finally got the good version working... Thanks for the help Zun, I got there in the end though!

zunath (approved):

PR #1863, opened 2024-08-12
Leadership aura fix

taelon875:
Now applies correct player movement when leaving aura.


SWLOR.Game.Server/Feature/StatusEffectDefinition/AuraStatusEffectDefinition.cs
```diff
@@ -53,6 +53,7 @@ private void Charge()
                 .RemoveAction((target, data) =>
                 {
                     RemoveEffectByTag(target, EffectTag);
+                    DelayCommand(0.5f,() => Stat.ApplyPlayerMovementRate(target));
```
zunath, line 56: Why is a delay needed here?

taelon875: It was needed to make the movement apply but I'm just going to shelf this for now and revisit at another time and try to find a way to make it would without running a delay after every time you get exit aura.

PR #1870, opened 2024-08-21
RebalanceShips&Bombs

taelon875:
Rebalanced New Ships
Rebalanced Space "Bosses"
Proton Damage Scales Better
Hypercapsule no longer recharges hostile targets.
Ammo More Missiles Per Craft as well as 2 new recipes for a T2 and T4 missile craft.

https://docs.google.com/spreadsheets/d/15cB79yZHryM-Ln-TZzG71z0A8Fp4ifjellofRfSvy4w/edit?gid=867387750#gid=867387750


SWLOR.Game.Server/Feature/MigrationDefinition/ServerMigration/_15_ShipRebalance.cs
```diff
@@ -0,0 +1,118 @@
+﻿using System;
```
zunath, line 1: image

zunath: The logic in this file should be put into ServerMigrationBase. Put it under a new function, then call that in this migration. Also update migration 14 so we aren't duplicating code

SWLOR.Game.Server/Service/CraftService/RecipeType.cs
```diff
@@ -2387,9 +2387,9 @@ public enum RecipeType
         InstructionShielding4 = 4686,
         InstructionInfusion1 = 4687,
         InstructionInfusion2 = 4688,
-        Missile3 = 4689,
-        Missile10 = 4690,
-        Missile25 = 4691,
+        Missile5 = 4689,
```
zunath, line 2390: You can't rename them once we've released. It'll cause an error for anyone who has already crafted these. You need to set up a migration script for it.

taelon875: I mean it would be easier to just leave the old names, they just are backend pointing at them i just figured you would want them to  match the quantities for the craft.

SWLOR.Game.Server/Feature/ShipModuleDefinition/HypermatterInjectorModuleDefinition.cs
```diff
@@ -51,7 +51,7 @@ private void HypermatterInjector(string tag, string name, string shortName, stri
                 })
                 .ActivatedAction((activator, activatorShipStatus, target, targetShipStatus, moduleBonus) =>
                 {
-                    if (!GetIsObjectValid(target))
+                    if (!GetIsObjectValid(target) || GetIsEnemy(target, activator))
```
zunath, line 54: What's the use case here?

SWLOR.Game.Server/Service/CraftService/RecipeType.cs
zunath: Alternatively keep the existing recipe and add these new options, adjusting levels & materials appropriately.

zunath (changes requested):
Requesting changes

SWLOR.Game.Server/Service/CraftService/RecipeType.cs
zunath: > I mean it would be easier to just leave the old names, they just are backend pointing at them i just figured you would want them to match the quantities for the craft.

Yea I do want them to match but you can't make that change without breaking existing character files. So either a migration script or introduce them as new recipes

taelon875: Yeah i'll just revert the originals in the recipe type and the craft and just add the new ones. Less headaches!

zunath:
Also, I updated the Starships tab but I need the specific changes on the Engineering Recipes page. There's a chance your version is missing some of the new entries. What am I copying over?

taelon875:
Missile Recipes on lines 1388-1392 on mine and then the quantities of Proton Bombs and assault concussion missiles

SWLOR.Game.Server/Feature/ShipModuleDefinition/HypermatterInjectorModuleDefinition.cs
taelon875: It is what makes the game automatically fire the device on self if target is hostile. I don't question how it works. it just does.

zunath: If other players are targeted in space I don't think it automatically hostiles them. In that case I don't think this will work as intended when targeting players.

I'm questioning whether the if statement is needed at all. If the expectation is that no matter what your target is, the hypermatter injector should be used then I'd say remove the if statement altogether. It seems unnecessary based on what I know.

taelon875: I mean the shield and hull use the same code. I just based it off those and they work fine. This is just checking to see if the target is hostile and if so it fires on self. If its another player or friendly it fires on them.

zunath:
Proton Bomb quantities are 5 in both sheets. Is that right?

SWLOR.Game.Server/Feature/RecipeDefinition/EngineeringRecipeDefinition/StarshipRecipes.cs
```diff
@@ -48,12 +48,12 @@ private void Tier1()
                 .Component("fiberp_ruined", 3)
                 .Component("elec_ruined", 3);
 
-            // Ship Missiles x3
-            _builder.Create(RecipeType.Missile3, SkillType.Engineering)
+            // Ship Missiles x5
```
zunath, line 51: I should have caught this on the initial submission, but these ammo recipes shouldn't be in the StarshipRecipes.cs class. I think they should be moved into either the ModuleRecipes.cs file or their own StarshipAmmo.cs class.

taelon875:
Yeah forgot the proton bombs were already at 5 on sheet, just set to single craft from code.

SWLOR.Game.Server/Feature/RecipeDefinition/EngineeringRecipeDefinition/StarshipRecipes.cs
taelon875: Wont moving them over directly cause issues to crafted things though?

SWLOR.Game.Server/Feature/ShipModuleDefinition/HypermatterInjectorModuleDefinition.cs
```diff
@@ -51,7 +51,7 @@ private void HypermatterInjector(string tag, string name, string shortName, stri
                 })
```
zunath, line 51: This was already here, but there's a typo for the short name on line 21. "FUel" -> "Fuel" Please correct this as part of this PR.

SWLOR.Game.Server/Feature/RecipeDefinition/EngineeringRecipeDefinition/StarshipRecipes.cs
zunath: The class files have no bearing on how the data gets stored. You can freely move them between the files - they're just a way of grouping recipes together. If I need to find the recipes for missiles, one of the last places I'll think to look is in the StarshipRecipes file. It's just a copy/paste job of moving it out of this file and into the StarshipModules or StarshipAmmo class, whichever you end up doing. I don't have a preference - but it shouldn't be in here with the starship recipes

SWLOR.Game.Server/Feature/ShipModuleDefinition/HypermatterInjectorModuleDefinition.cs
zunath: Ok I see now. It's fine to leave this

SWLOR.Game.Server/Feature/RecipeDefinition/EngineeringRecipeDefinition/StarshipRecipes.cs
taelon875: Okay no problems moving it, I'll make it a StarshipAmmo, it makes the most sense being as we now have what 4 ammo types for ships.

zunath:
More stuff we discussed regarding the player market additions:

1. Add an IsStarshipAmmo method to Space.cs. Have the rules here for the resref check

2. Add an entry to MarketCategoryType.cs for Starship Ammo

3. Add the logic to call Space.IsStarshipAmmo() to the end of GetItemMarketCategory()

zunath (approved):

PR #1877, opened 2024-08-25
Legion fix

Scorchys12:
Fixed transition from Lake to Legion base.


zunath:
I don't see how this fixes the problem. All that changed is something with the tile border. Make sure you actually fixed it and it was committed correctly.

PR #1891, opened 2025-02-10
Korriban areas QOL + quests updates and snow world prefabs

Scorchys12:


taelon875 (changes requested):
The Apprentice robe needs to be setup correctly, it doesn't have a armor type associated with it currently

PR #1902, opened 2025-03-13
East Map Push

Aephyros:


Aephyros:
Untitled

taelon875 (changes requested):
Lots of file changes here, if I had to assume i would guess that your module wasn't up to date so it looks like it is removing some previous areas that were modified/changed.  Please go through and make sure to only send of the files of your areas.

PR #1922, opened 2025-08-03
Add check for possessed NPCs when sending translated messages

zunath:
## Summary by CodeRabbit

* **Bug Fixes**
  * Improved language translation handling so that DM-possessed creatures always understand text, ensuring consistent behavior for DMs.


zunath:
@coderabbitai - Per this documentation: https://lexicon.nwn.wiki/index.php/GetIsDM  the assumption that GetIsDM will return TRUE is incorrect. It will return false if the NPC is possessed by a DM.

PR #1983, opened 2026-05-02
Feature/improved stuck command

zunath:


zunath:
@coderabbitai - As a result of this change, add a rule which checks to make sure area files contain a Stuck Waypoint on all future PR submissions. Warn the contributor if they forget to add this, but don't treat it as an error (some areas, like prefabs do not need this waypoint).

PR #1987, opened 2026-05-02
Address discrepancy where Improved Power Supply I-V granted one less …

zunath:
…Module Enhancement than expected. Update all store instances.


zunath:
@coderabbitai - Please review if rate limit is over.

zunath:
@coderabbitai - Need that review, buddy.

PR #1993, opened 2026-05-04
Due to item stacking issues, rely on a true/false flag instead of the…

zunath:
… individual player's ID. Don't do this for fish due to stacking issues.


zunath:
@coderabbitai - full review

PR #2075, opened 2026-07-13
Veles Road Clipping Fix

Aephyros:
Increased the Z axis of every road placeable by .01 Goodbye, clipping, we knew you well.

## Summary by CodeRabbit

* **Bug Fixes**
  * Updated the Veles exterior area data for improved consistency.
  * Adjusted the positioning and elevation of several exterior elements.
  * Refreshed module identification data to ensure the updated area loads correctly.


zunath:
Reviewed the CodeRabbit findings:

**Mod_ID encoding (valid, confirmed via byte-level diff of `Module/ifo/module.ifo.json`):** on `master` the Mod_ID's first 4 bytes are the properly UTF-8-encoded replacement character (U+FFFD, i.e. `EF BF BD` repeated 4 times). This PR's branch overwrites those same 4 bytes with raw `0xFF` bytes, which are not valid UTF-8 - the JSON file is now byte-invalid text at that spot. This PR doesn't intentionally touch Mod_ID at all (it's a road-clipping Z-fix), so the corruption looks like a side effect of the toolset resave that produced this diff. Fix: revert just that one field's leading 4 bytes back to master's value (the trailing `...` bytes are unchanged/fine).

**Missing STUCK_WAYPOINT:** not something to fix in this PR. `veles_exterior.git.json` already lacks a `STUCK_WAYPOINT` on `master`, and this PR's diff doesn't touch waypoints at all - it's a pre-existing area-design gap unrelated to the clipping fix, and adding one requires picking a real in-world location, which is a separate design decision.

PR #2077, opened 2026-07-13
Add /hpset in-chat HP tracker with above-the-head name-plate bar

garcia-pedro-hr:
## Summary
Adds a lightweight, **narrative** HP tracker for creatures (PCs and NPCs), shown as a colored bar in the creature's **name-plate above its model**. It's a pure bookkeeping/display tool — completely separate from the engine's real combat HP (e.g. track `4/5` without ever damaging or healing the creature).

## Commands
All are click-to-target. Regular players may only target **themselves**; DMs/Admins may target **any** creature (gated via `Authorization.GetAuthorizationLevel`, matching the existing DM commands).

| Command | Alias | Behavior |
|---|---|---|
| `/hpset <current> [max]` | `/sethp` | Create/replace a tracker. One arg starts full; two args set `current/max` (e.g. `/hpset 3 8` → 3/8). |
| `/hpinc [amount]` | `/inchp` | Raise current HP (default +1), capped at max. |
| `/hpdec [amount]` | `/dechp` | Lower current HP (default -1), floored at 0. |
| `/hpdel` | `/delhp` | Remove the tracker. |

`/hpset`, `/hpinc`, and `/hpdec` broadcast the new value to everyone in the target's area; `/hpdel` feedback is private to the caller.

## Implementation
- **`Service/HPTracker.cs`** — an in-memory `Dictionary<uint,(Current,Max)>` registry (transient, Enmity-style lifecycle). Renders the bar per-observer via NWNX `PlayerPlugin.SetCreatureNameOverride` (works on PCs, unlike `SetName`). Observers are enumerated with `GetFirstPC`/`GetNextPC` filtered to the creature's area (this includes DM clients, unlike `Area.GetPlayersInArea`). Overrides are re-applied on `OnAreaEnter` (both when a viewer arrives and when a tracked creature moves) and purged on `OnObjectDestroyed` and player logout (`OnModuleExit`). The bar is a 10-segment green→yellow→red gradient via `ColorToken`.
- **`Feature/ChatCommandDefinition/HPTrackerChatCommand.cs`** — the four commands (auto-discovered `IChatCommandListDefinition`), argument validation, the self-vs-staff gate, and the area broadcast.

No existing files are modified, and the tracker never touches real hit points.

## Notes
- The bar renders as thin colored strokes rather than solid blocks — NWN's floating-name font has no Unicode block glyphs (verified in-game by probing `█ ▓ ■ ▬ ▮ #`). A true graphical bar would require screen-space NUI, which can't be anchored in world-space above a model.
- Trackers are in-memory and cleared on server restart / tracked-PC logout (matches the "temporary" intent).

## Testing
Verified live on a local server: the bar appears above an NPC and updates live on inc/dec for observing DMs/players; broadcasts fire to the area; argument validation, permission gating, and cleanup behave as described; and the creature's real combat HP is untouched throughout.

image
image
image
image

## Summary by CodeRabbit

* **New Features**
  * Added temporary HP trackers with visible current/max values and a color-coded HP bar.
  * Added chat commands to set, increase, decrease, and remove tracked HP.
  * HP updates are displayed to nearby players and DMs.
  * Trackers automatically refresh when creatures or observers change areas.
* **Bug Fixes**
  * Removed stale HP bars when tracked creatures are destroyed or players leave.

### DIFF
--- a/SWLOR.Game.Server/Feature/ChatCommandDefinition/HPTrackerChatCommand.cs
+++ b/SWLOR.Game.Server/Feature/ChatCommandDefinition/HPTrackerChatCommand.cs
@@ -74,8 +74,7 @@ namespace SWLOR.Game.Server.Feature.ChatCommandDefinition
                     }
 
                     HPTracker.Set(target, current, max);
-                    var (c, m) = HPTracker.Get(target);
-                    BroadcastNearby(target, ColorToken.Green($"{GetName(target)}'s HP tracker set to {c}/{m}."));
+                    BroadcastNearby(target, ColorToken.Green($"{GetName(target)}'s HP tracker set to {current}/{max}."));
                 });
         }
 
@@ -155,8 +154,8 @@ namespace SWLOR.Game.Server.Feature.ChatCommandDefinition
             }
 
             HPTracker.Adjust(target, delta);
-            var (c, m) = HPTracker.Get(target);
-            BroadcastNearby(target, ColorToken.Green($"{GetName(target)}'s HP tracker is now {c}/{m}."));
+            var (current, max) = HPTracker.Get(target);
+            BroadcastNearby(target, ColorToken.Green($"{GetName(target)}'s HP tracker is now {current}/{max}."));
         }
 
         /// <summary>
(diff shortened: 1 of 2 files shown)


garcia-pedro-hr:
Superseded by #2078. Reworked onto `feature/combat-upgrade` and scoped to NPC-only so the HP tracker stays fully independent of the player recognition/disguise system (which owns PC name display via NWNX Rename) — no piggyback, no name-plate/chat conflict.

PR #2078, opened 2026-07-13
Add /hpset NPC HP tracker with above-the-head name-plate bar

garcia-pedro-hr:
Rebased onto `feature/combat-upgrade` and reworked from #2077 (which targeted `master`) to stay independent of the player recognition/disguise system on this branch.

## Summary
A temporary, **narrative** HP tracker shown as a colored bar in an **NPC's** name-plate (above its model), completely separate from the engine's real combat HP — e.g. track `4/5` on a scene NPC without ever damaging or healing it.

## Scope: NPC-only, by design
The recognition/disguise system (`PlayerName` + NWNX Rename) owns **PC** displayed names. Rather than piggyback on that (which would put the bar inside the Rename override — and Rename has no name-plate/chat split, so it would leak into a tracked PC's chat/tooltip), this feature is **NPC-only**: it renders via a different channel (`PlayerPlugin.SetCreatureNameOverride`) that never touches PCs. The two systems therefore stay fully independent — no clobbering, no chat leak, no disguise conflict.

- Targets are restricted to non-PC, non-DM creatures; PC/DM targets are rejected with a clear message.
- Commands are DM/Admin-only (a scene-management tool). They register with `AuthorizationLevel.All` so non-staff get a real message rather than the generic "invalid command."

## Commands (click-to-target an NPC)
| Command | Alias | Behavior |
|---|---|---|
| `/hpset <current> [max]` | `/sethp` | Create/replace a tracker. One arg starts full; two args set `current/max` (e.g. `/hpset 3 8` → 3/8). |
| `/hpinc [amount]` | `/inchp` | Raise current HP (default +1), capped at max. |
| `/hpdec [amount]` | `/dechp` | Lower current HP (default -1), floored at 0. |
| `/hpdel` | `/delhp` | Remove the tracker. |

`/hpset`, `/hpinc`, `/hpdec` broadcast the new value to everyone in the target's area; `/hpdel` is private to the caller.

## Implementation
- **`Service/HPTracker.cs`** — in-memory `Dictionary<uint,(Current,Max)>` registry (transient). Renders the bar per-observer via `PlayerPlugin.SetCreatureNameOverride` to players/DMs in the NPC's area, re-applied on `OnAreaEnter`, cleared on `OnObjectDestroyed` (and player logout), clearing the override before dropping the tracker so a reused object id can't inherit a ghost bar. 10-segment green→yellow→red gradient via `ColorToken`.
- **`Feature/ChatCommandDefinition/HPTrackerChatCommand.cs`** — the four commands (auto-discovered `IChatCommandListDefinition`), argument validation, the NPC-only + DM/Admin gate, and the area broadcast.

No existing files are modified; the tracker never touches real hit points.

## Notes
- The bar renders as thin colored strokes, not solid blocks — NWN's name-plate font has no Unicode block glyphs (verified in-game by probing `█ ▓ ■ ▬ ▮ #`).
- Trackers are in-memory and cleared on server restart (matches the "temporary" intent).

## Open question — PC support
If PC HP bars are wanted later, they'd need to compose into `PlayerName` (there's no other persistent above-the-model surface), which means the bar rides the Rename name override and would also appear in the PC's chat/tooltip name. Happy to add that (with or without a chat-event swap to strip it during messages) if desired — left out here to keep this change independent and low-risk.

### DIFF
--- a/SWLOR.Game.Server/Feature/ChatCommandDefinition/HPTrackerChatCommand.cs
+++ b/SWLOR.Game.Server/Feature/ChatCommandDefinition/HPTrackerChatCommand.cs
@@ -1,0 +1,209 @@
+using System.Collections.Generic;
+using SWLOR.Game.Server.Enumeration;
+using SWLOR.Game.Server.Service;
+using SWLOR.Game.Server.Service.ChatCommandService;
+using SWLOR.NWN.API.NWScript.Enum;
+
+namespace SWLOR.Game.Server.Feature.ChatCommandDefinition
+{
+    /// <summary>
+    /// Chat commands for a temporary, narrative HP tracker shown as a bar above an NPC.
+    /// See <see cref="HPTracker"/> for the display/lifecycle logic. These commands never
+    /// affect a creature's real combat hit points.
+    ///
+    /// All four require the caller to click a creature target. Trackers are NPC-only and
+    /// DM/Admin-only: player-character name display is owned by the recognition/disguise
+    /// system (<see cref="PlayerName"/>), so keeping HP trackers off PCs guarantees the two
+    /// systems remain independent and never fight over the same displayed name.
+    /// </summary>
+    public class HPTrackerChatCommand : IChatCommandListDefinition
+    {
+        private readonly ChatCommandBuilder _builder = new ChatCommandBuilder();
+
+        public Dictionary<string, ChatCommandDetail> BuildChatCommands()
+        {
+            SetHP();
+            IncreaseHP();
+            DecreaseHP();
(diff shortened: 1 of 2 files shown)


zunath:
Feedback from QM/DM team:

1. The colored bar isn't necessary and can get really busy considering the disguise system coming in. Recommend removal
2. It would be best to add something like /eventhp which opens a NUI. The NUI should let you target a player, then set/increase/decrease HP. This will be easier to use than chat commands.
3. Do a final confirmation that this is a temporary change and doesn't persist.

PR #2082, opened 2026-07-15
Add HP Tracker: temporary narrative HP tracking (/hptracker window + /hpset commands)

garcia-pedro-hr:
## HP Tracker — a temporary, narrative HP-tracking tool

A DM/GM and player tool for tracking **narrative** hit points on any creature. It is
display-only: it **never applies damage or healing and never touches a creature's real
combat HP**, and nothing is persisted — trackers clear on removal, creature death/destruction,
player logout, and server restart.

### Opening it
**`/hptracker`** opens the HP Tracker window (below). All tracker management — add, adjust,
remove — happens inside the window itself; there are no other commands.

### HP Tracker window
- Lists every tracked creature in your area, one row each: **name · HP-% bar (green→yellow→red) · `current/max` · `Loc` · `−` · `+` · `x`**.
- **Add (target):** enter an HP value and target a creature to start tracking it (rejects non-creatures, unauthorized targets, and already-tracked creatures).
- **`−` / `+`** adjust the tracked HP (clamped to `0..max`); **`x`** removes the tracker.
- **`Loc`** glows the creature in the world **only you can see** (a per-observer looping VFX), colored by its HP (green→yellow→red) and recoloring live as HP changes; click again to clear.
- Live-refreshes on any HP change and on area transitions (a tracked creature or a viewer moving between areas) — no heartbeat polling; rows stay in a stable order. Resizable/collapsible with per-player saved geometry.

window

hpm
hpf
hpl

### Permissions
Staff (DM / DM-possessed / Admin) can manage any creature; normal players can manage only their
own PC — enforced **server-side** in the handlers, not merely via disabled buttons. Locating is
always allowed. The window opens correctly for DM-possessed creatures, and a bare unpossessed DM
gets a helpful message rather than a silent no-op.

### Scope
Purely additive (7 new files, ~1 line touched elsewhere):
- `Service/HPTracker.cs` — pure in-memory state (unit-test friendly); all presentation lives in the ViewModel.
- `Feature/HPTrackerWindow.cs` — refresh (on tracker changes and area transitions), object/player lifecycle cleanup, and the shared permission rules.
- `Feature/GuiDefinition/HPTrackerDefinition.cs` + `ViewModel/HPTrackerViewModel.cs` + `RefreshEvent/HPTrackerRefreshEvent.cs` — the NUI window.
- `Feature/ChatCommandDefinition/HPTrackerChatCommand.cs` — the `/hptracker` opener.
- The only change to an existing file is a single new `GuiWindowType` enum value.

## Summary by CodeRabbit

- **New Features**
  - Added an HP Tracker window accessible through `/hptracker`.
  - Track creatures’ current and maximum HP within the viewer’s area.
  - Add, increase, decrease, remove, and locate tracked creatures.
  - HP bars display current status with color-coded health levels.
  - Tracker data refreshes when HP changes, creatures move, or tracked creatures leave play.


SWLOR.Game.Server/Feature/ChatCommandDefinition/HPTrackerChatCommand.cs
```diff
@@ -0,0 +1,229 @@
...
+        {
+            if (!CanTarget(user, target)) return;
+
+            if (!HPTracker.Has(target))
+            {
+                SendMessageToPC(user, ColorToken.Red($"{GetName(target)} does not have an HP tracker. Use /hpset first."));
+                return;
+            }
+
+            HPTracker.Adjust(target, delta);
```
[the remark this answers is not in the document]
garcia-pedro-hr: Good catch. Fixed in `f177b2fc6`: `HPTracker.Adjust` now widens the running total to `long` before clamping to `[0, Max]`, so a tracker set near `int.MaxValue` (e.g. via `/hpset`) cannot overflow on a later `/hpinc` or GUI increment. Applied the root fix in `Adjust` so both the command and GUI increment paths are covered.

zunath:
@coderabbitai full review

SWLOR.Game.Server/Feature/GuiDefinition/ViewModel/HPTrackerViewModel.cs
```diff
@@ -0,0 +1,260 @@
...
+        /// Applies this viewer's locate-glow to a creature, colored by its tracked HP (green -> yellow ->
+        /// red), and records it so <see cref="Rebuild"/> can recolor or clear it later.
+        /// </summary>
+        private void ApplyHighlight(uint creature)
+        {
+            var aura = AuraForCreature(creature);
+            PlayerPlugin.ApplyLoopingVisualEffectToObject(Player, creature, aura);
+            _highlightedCreature = creature;
+            _highlightAura = aura;
+        }
```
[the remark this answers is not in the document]
garcia-pedro-hr: Good catch. Fixed in `c1a2fb7b9`: `ApplyHighlight` now guards `GetIsObjectValid(creature)` before applying the looping VFX, mirroring `ClearHighlight`. On an invalid target it returns early, leaving `_highlightedCreature` at `OBJECT_INVALID` (set by the preceding `ClearHighlight()` in `OnClickName`), so the glow state stays clean.

zunath:
@coderabbitai - full review

SWLOR.Game.Server/Feature/HPTrackerWindow.cs
```diff
@@ -0,0 +1,84 @@
...
+            RefreshOpenWindows();
+        }
+
+        /// <summary>A destroyed creature drops its tracker (its object id can be reused).</summary>
+        [NWNEventHandler(ScriptName.OnObjectDestroyed)]
+        public static void RemoveTrackerOnDestroyed()
+        {
+            if (HPTracker.Remove(OBJECT_SELF))
+                RefreshOpenWindows();
+        }
```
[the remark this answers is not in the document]
garcia-pedro-hr: Good catch. Fixed in `1de1a6606`: added `RemoveTrackerOnDeath` on `OnCreatureDeathAfter`, alongside the existing `OnObjectDestroyed` cleanup. A corpse can stay valid-but-dead for a long time before destruction fires, so death is the right moment to drop the tracker and refresh open windows. Also updated the `HPTracker` class doc comment to list death as a removal trigger.

SWLOR.Game.Server/Service/HPTracker.cs
```diff
@@ -0,0 +1,115 @@
...
+        public static (int Current, int Max) Get(uint creature)
+        {
+            var state = _trackers[creature];
+            return (state.Current, state.Max);
+        }
+
+        /// <summary>Tracked creatures in the same area as the viewer (the "in range" rule for the window).</summary>
+        public static List<uint> GetTrackedInArea(uint viewer)
+        {
+            var area = GetArea(viewer);
```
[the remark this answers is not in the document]
garcia-pedro-hr: Good catch. Fixed in `1de1a6606`: `GetTrackedInArea` now returns an empty list when `GetArea(viewer)` is invalid (e.g. dead in Limbo or mid-transition), so an arealess viewer no longer matches every other arealess tracked creature and exposes unrelated trackers. The same commit also caches each `GetName` once before sorting, which addresses the separate O(N log N) interop nitpick on this method.

garcia-pedro-hr:
Pushed `1de1a6606` addressing the outstanding review items. Summary of what changed and what I'm intentionally leaving as-is:

**Applied**
- **Death cleanup** — `RemoveTrackerOnDeath` on `OnCreatureDeathAfter` (a corpse can linger valid-but-dead before `OnObjectDestroyed` fires).
- **Invalid viewer area guard** — `GetTrackedInArea` returns an empty list when the viewer's area is invalid, so arealess viewers don't match unrelated trackers.
- **Sort perf** — `GetTrackedInArea` caches each `GetName` once before sorting instead of calling it O(N log N) times per heartbeat.
- **Serilog audit logging** — staff-driven mutations now log via `Log.Write(LogGroup.DM, ...)` with actor / target / before-after, at both the chat commands (`/hpset`, `/hpdel`, `/hpinc`, `/hpdec`) and the window buttons (Add / x / + / -). Player self-service (managing your own narrative HP) is intentionally not logged, to keep the audit trail signal-only.
- **Formatting** — split the `AddLabel().SetText("")` fluent chain onto separate lines in `HpTrackerDefinition`.

**Intentionally not changing**
- **`CanTarget` reuse of `IsTrackableTarget`/`CanManage`** — the permission rule (`IsStaff`) is already shared via `HpTrackerWindow`; the only remaining overlap is the trackable-target check. Collapsing it into `IsTrackableTarget` would drop the distinct "can't place an HP tracker on a DM" message, which is friendlier than the merged wording. Keeping the explicit messages.
- **Unit tests for `HPTracker`** — there's no test project in the repo today, so this isn't the "cheap to add" it would be elsewhere. Deferring rather than standing one up in this PR.
- **`HP` → `Hp` rename** — cosmetic with a wide reference surface; I'd rather not churn the diff mid-review. Happy to do it as a dedicated follow-up if unifying the casing is wanted.

SWLOR.Game.Server/Feature/GuiDefinition/ViewModel/HpTrackerViewModel.cs
```diff
@@ -0,0 +1,281 @@
...
+                {
+                    SendMessageToPC(Player, ColorToken.Red($"{GetName(creature)} is already being tracked."));
+                    return;
+                }
+
+                HPTracker.Set(creature, hp, hp);
+                HpTrackerWindow.RefreshOpenWindows();
+
+                if (HpTrackerWindow.IsStaff(Player))
+                    Log.Write(LogGroup.DM, $"HP tracker set (window): Actor={GetName(Player)}, Target={GetName(creature)}, After={hp}/{hp}");
```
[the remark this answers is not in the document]
garcia-pedro-hr: Good call. Fixed in `b8a7650b3`: the actor identity in every DM audit line now carries the CD key — `Actor={GetName(x)} ({GetPCPublicCDKey(x)})` — matching the `/restartserver` audit precedent. Applied it to all three window sites flagged here plus the three chat-command sites (`/hpset`, `/hpdel`, `/hpinc` / `/hpdec`), so both audit paths log the same raw player identity.

SWLOR.Game.Server/Feature/ChatCommandDefinition/HPTrackerChatCommand.cs
```diff
@@ -0,0 +1,242 @@
...
+                    {
+                        SendMessageToPC(user, ColorToken.Red("The HP Tracker window can't open for an unpossessed DM. Possess a creature first, or use a player character."));
+                        return;
+                    }
+
+                    Gui.TogglePlayerWindow(player, GuiWindowType.HpTracker, null, OBJECT_INVALID, uiTarget);
+                });
+        }
+
+        private void SetHP()
```
zunath: Since we're going with a UI approach, I don't think these commands are needed anymore.

SWLOR.Game.Server/Feature/GuiDefinition/RefreshEvent/HpTrackerRefreshEvent.cs
```diff
@@ -0,0 +1,11 @@
+using SWLOR.Game.Server.Service.GuiService;
+
+namespace SWLOR.Game.Server.Feature.GuiDefinition.RefreshEvent
+{
+    /// <summary>
+    /// Signals the HP Tracker window to rebuild its list (HP changed, or creatures moved in/out of range).
+    /// </summary>
+    public class HpTrackerRefreshEvent : IGuiRefreshEvent
```
zunath: Can the PlayerStatusRefreshEvent be reused?

If not, please change this from Hp to HP since it's an acronym.

SWLOR.Game.Server/Feature/GuiDefinition/HpTrackerDefinition.cs
```diff
@@ -0,0 +1,140 @@
...
+{
+    /// <summary>
+    /// The "HP Tracker" scene window (opened with /hptracker). A header row (HP value + Add-by-target),
+    /// then one list row per tracked creature in range: the creature name (fills the row), a colored HP
+    /// bar, "cur/max", a Locate button (glow only the viewer sees), and per-row -/+/x buttons. Structure
+    /// mirrors the proven Bank/MarketListing windows (resizable; a variable-width name label fills the row;
+    /// every action is a fixed-width button so nothing collapses). Never touches real combat HP. See
+    /// <see cref="ViewModel.HpTrackerViewModel"/>.
+    /// </summary>
+    public class HpTrackerDefinition : IGuiWindowDefinition
```
zunath: Hp -> HP

SWLOR.Game.Server/Feature/HpTrackerWindow.cs
```diff
@@ -0,0 +1,95 @@
...
+            {
+                if (!Gui.IsWindowOpen(player, GuiWindowType.HpTracker))
+                    continue;
+
+                var viewModel = Gui.GetPlayerWindow(player, GuiWindowType.HpTracker).ViewModel;
+                (viewModel as IGuiRefreshable<HpTrackerRefreshEvent>)?.Refresh(new HpTrackerRefreshEvent());
+            }
+        }
+
+        [NWNEventHandler(ScriptName.OnSwlorHeartbeat)]
```
zunath: Please avoid using heartbeats.

SWLOR.Game.Server/Service/GuiService/GuiWindowType.cs
```diff
@@ -60,6 +60,7 @@ public enum GuiWindowType
         MusicPicker = 55,
         Dice = 56,
         GuildTasks = 57,
+        HpTracker = 58,
```
zunath: Hp -> HP if you keep this.

SWLOR.Game.Server/Service/HPTracker.cs
```diff
@@ -0,0 +1,127 @@
...
+
+            var result = new List<uint>(inArea.Count);
+            foreach (var item in inArea)
+                result.Add(item.Id);
+
+            return result;
+        }
+
+        /// <summary>Bar fill 0..1 for the given current/max (pure).</summary>
+        public static float GetProgress(int current, int max)
```
zunath: This stuff is UI-based and shouldn't live in the service. It probably should be placed within the ViewModel

SWLOR.Game.Server/Service/HPTracker.cs
```diff
@@ -0,0 +1,127 @@
...
+        {
+            if (max <= 0) return 0f;
+            var ratio = (float)current / max;
+            if (ratio < 0f) return 0f;
+            if (ratio > 1f) return 1f;
+            return ratio;
+        }
+
+        /// <summary>Bar color, green (full) -> yellow (half) -> red (empty) by ratio (pure).</summary>
+        public static GuiColor GetBarColor(int current, int max)
```
zunath: This stuff is UI-based and shouldn't live in the service. It probably should be placed within the ViewModel

zunath (changes requested):
Requesting changes

SWLOR.Game.Server/Feature/ChatCommandDefinition/HPTrackerChatCommand.cs
garcia-pedro-hr: Done in `c08525d12` — removed `/hpset`, `/hpinc`, `/hpdec`, `/hpdel` and all their helpers; the window is now the only management surface. I kept `/hptracker` itself as the opener, since the window still needs an entry point (same pattern as `/dice`).

SWLOR.Game.Server/Feature/GuiDefinition/RefreshEvent/HpTrackerRefreshEvent.cs
garcia-pedro-hr: I looked at reusing it but went with the rename instead (`HPTrackerRefreshEvent`, in `c08525d12`). `PlayerStatusRefreshEvent` is the real-vitals contract — a required `StatType` (HP/FP/STM/Shield/Hull/Capacitor) consumed by `PlayerStatusViewModel` and `CharacterSheetViewModel` — while this tracker is narrative-only and never reads real vitals. Sharing the type would rebuild the HP Tracker on every combat vitals publish and, conversely, refresh the player status windows on tracker edits. I documented that distinction on the event class so it doesn't need re-deriving later.

SWLOR.Game.Server/Feature/GuiDefinition/HpTrackerDefinition.cs
garcia-pedro-hr: Done in `c08525d12` — `HPTrackerDefinition` (class + file), with the same casing applied to the ViewModel, the window coordinator, and the bound properties (`AddHPText`, `HPProgresses`, `HPColors`, `HPTexts`).

SWLOR.Game.Server/Feature/HpTrackerWindow.cs
garcia-pedro-hr: Removed in `c08525d12`. The list only changes when a tracked creature or a viewer crosses areas, so the heartbeat is replaced with `area_enter`/`area_exit` handlers that refresh exactly on those transitions (with a guard for DM-possessed movers, whose window lives under the master DM). Every other refresh was already event-driven from explicit tracker changes (add/adjust/remove, death, destruction, logout).

SWLOR.Game.Server/Service/GuiService/GuiWindowType.cs
garcia-pedro-hr: Kept — the window needs its own type for open/close state and saved geometry — and renamed to `HPTracker` in `c08525d12`.

SWLOR.Game.Server/Service/HPTracker.cs
garcia-pedro-hr: Agreed — `GetProgress` and `GetBarColor` moved into `HPTrackerViewModel` as private helpers in `c08525d12`. The `HPTracker` service is now pure state and no longer references any GUI type.

SWLOR.Game.Server/Service/HPTracker.cs
garcia-pedro-hr: Moved into `HPTrackerViewModel` together with `GetProgress` in `c08525d12` (same fix as the comment above); the service no longer imports `GuiService.Component`.

garcia-pedro-hr:
@coderabbitai Good catch on the bare-DM guard. Fixed in `33e40d0d0`: `GetIsPC()` returns true for a player-controlled DM avatar (which is why the codebase idiom is `!GetIsPC(x) || GetIsDM(x)`), so the old `!GetIsPC(player)` condition never fired and an unpossessed DM fell through to `Gui.TogglePlayerWindow()`'s silent no-op. The guard now checks `GetIsDM(player)` explicitly (with the same `uiTarget` condition preserved, so the DM-possessed flow still opens the window on the possessed creature's client), and the guidance message actually sends.

zunath:
@codex review

PR #2092, opened 2026-07-18
Let Espionage perks extend the Disguises system

zunath:
Wires the Espionage Infiltrator tree into the Disguises system, which until now had no perk, stat, or skill coupling at all.

Depends on zunath/SWLOR_Haks#347 for the feat rows, TLK entries, and icons. The submodule pointer here references that commit.

## Perks

Both land in a new `PerkCategoryType.EspionageTradecraft` (60), shown as **Espionage - Tradecraft**. These are general identity tradecraft rather than infiltration or sabotage, so they get their own category instead of riding on an existing tree — and it leaves a home for future non-combat Espionage utility perks.

| Perk | Ranks | SP | Espionage | Effect |
| --- | --- | --- | --- | --- |
| False Identities I-III | 3 | 2/3/4 | 10/28/44 | `AdditionalDisguiseSlots` +1/+2/+3, so 2/3/4 slots total |
| Cover Story I-II | 2 | 3/3 | 20/40 | `DisguiseSwapCooldownReductionPercent` 40/70, so 18 min / 9 min |

## Implementation

Both are modeled as `StatType` adjustments read by the Disguise service, per the stat-driven rule in AGENTS.md — no perk-type checks anywhere in `Disguise.cs`, and a test asserts that stays true. This mirrors `Traps.GetTrapCapacity` consuming `AdditionalTrapCapacity`.

- `GetDisguiseSlotLimit` now takes the creature and adds `Stat.GetStatAdjustment(player, StatType.AdditionalDisguiseSlots)` on top of the existing per-player DB grant. `Player.DefaultDisguiseSlotLimit` stays 1 and remains the floor.
- The activation delay resolves through `GetActivationDelay(player)`, clamped to 0-100% and floored at a new `MinimumActivationDelayMinutes = 5` so stacked reduction sources can never remove the delay. Identity-strobing would let a player dodge social consequence faster than observers can react, since each disguise is its own `disguise:<id>` key in the recognition system.
- Slot math and delay math are factored into pure `CalculateDisguiseSlotLimit` / `CalculateActivationDelay` functions so they are testable without a live creature.
- The delay message previously hardcoded `ActivationDelayMinutes`, so a perked player would have been told "there is a 30-minute delay" while actually waiting 9. It now reports the effective delay.

## Deliberately unchanged

Deactivating a disguise is still instant and still does not refund a running activation timer. Confirmed as intended: the timer models a 30-minute commitment to an identity, and briefly showing your real face spends it.

## Tests

`SWLOR.Game.Server.Tests/Perks/EspionageDisguisePerkTests.cs` covers rank prices/requirements/stat values, feat granted on rank 1 only, slot stacking (including on top of an admin grant and the negative-stat floor), the 30/18/9 delay bands, the 5-minute floor, and that the service reads stats rather than perks.

Full suite: 863 passed, 0 failed.

## Design Bible

The Espionage tab gets a new **Tradecraft** section inserted *above* the Infiltrator block, at rows 9-13 with its own `Total` subtotal (15 SP) at row 14 and a spacer at row 15:

| Style | SP | Perk | Skill Reqs. | Type | Description |
| --- | --- | --- | --- | --- | --- |
| Tradecraft | 2.0 | False Identities I | Espionage 10 | Trait | Increases the number of disguises you may keep on file to 2. |
| Tradecraft | 3.0 | Cover Story I | Espionage 20 | Trait | Reduces the delay between disguise activations by 40%. |
| Tradecraft | 3.0 | False Identities II | Espionage 28 | Trait | Increases the number of disguises you may keep on file to 3. |
| Tradecraft | 3.0 | Cover Story II | Espionage 40 | Trait | Reduces the delay between disguise activations by 70%. |
| Tradecraft | 4.0 | False Identities III | Espionage 44 | Trait | Increases the number of disguises you may keep on file to 4. |

Because the insert shifted rows 9+ down by 7, all three subtotal formulas were repointed along with their cached values: `SUM(B9:B26)` to `SUM(B16:B33)`, `SUM(B29:B46)` to `SUM(B36:B53)`, and the tab total `SUM(B27,B47)` to `SUM(B14,B34,B54)` = 135.

The workbook was edited at the zip/XML level rather than through a library that rewrites the whole file. Verified afterwards that only `xl/worksheets/sheet30.xml` and `xl/sharedStrings.xml` changed and the other 213 zip entries are byte-identical, so no cached formula values elsewhere were disturbed. `CombatUpgradeBibleWorkbookFormattingTests` and `NPCEnemyBalanceAuditTests` are green.

## One thing to flag

**`CombatUpgradeBibleImplementationReview.csv` churn is mostly not mine.** Most of its diff is the Espionage rows renumbering by 7, which is expected. But seven rows also flip a granted-feat label (e.g. `PropagationTrait` to `BreakerGuardTrait`). Root cause: `FeatType` has **13 pre-existing duplicate enum values** on `feature/combat-upgrade`, so `Enum.GetName` picks an arbitrary winner and adding any enum member reshuffles it. All verdicts stay PASS. Worth fixing separately — each duplicated pair also collapses onto a single `feat.2da` row, so one feat of each pair renders the other's icon and tooltip:

```
1403 DualEdgeRhythmTrait/IronGuardTrainingTrait   1431 PatientBreakerTrait/ButchersTempoTrait
1412 BloodWardTrait/RedirectingCounterTrait       1441 GuardDoctrineTrait/FirstStrikeTrait
1413 BloodPriceTrait/GuardReversalTrait           1442 CoveringPresenceTrait/VenaticRecoveryTrait
1414 RecklessPowerTrait/HypermetabolizeTrait      1443 CoverReversalTrait/CheapShotTrait
1424 OpeningReadTrait/DebilitateTrait             1444 GuardGripTrait/HobbledTrait
1425 PrecisionClawsTrait/InfectionTrait
1426 ExploitOpeningTrait/VenomTempoTrait
1430 BreakerGuardTrait/PropagationTrait
```

Note the Espionage tab is in `OutOfScopeTabs` for the sync tests, so the new Bible rows are not cross-checked against the perk definitions the way in-scope tabs are. `EspionageDisguisePerkTests` covers the code side directly instead.

Needs a hak rebuild and module repack on deploy.

## Summary by CodeRabbit

- **New Features**
  - Added the Espionage Tradecraft perk category.
  - Added False Identities (additional disguise slots) and Cover Story (reduced disguise activation cooldowns), including their progression and related traits.

- **Improvements**
  - Disguise slot limits and activation cooldowns now reflect the character’s applicable bonuses rather than a fixed baseline.
  - Activation cooldown messaging and confirmation prompts now show the effective computed delay.
  - Cooldowns are still protected by a minimum wait time, with large reductions floored and negative reductions ignored.


zunath:
@coderabbitai full review

zunath:
﻿Fixed in 5dcc215b3 — good catch, and it was worse than the two modals.

Grepping for the hardcoded value turned up a third surface you didn't flag: `DisguiseDefinition.cs:391` had the note label as a fixed `SetText("Activating starts a 30-minute cooldown…")`. Because that one lives in the window layout rather than the view model, it needed a `BindText` against a new `ActivationDelayNote` property refreshed alongside the slot usage, not just string interpolation.

All three now derive from `Disguise.GetActivationDelay(Player)` via a shared `GetActivationDelayMinutes()` helper, and the source-string assertions in `PlayerNameRecognitionTests` were updated to match.

I also added a regression test that asserts no disguise UI surface contains the literal `30-minute`, so the next surface someone adds can't quietly reintroduce it.

Full suite: 864 passed, 0 failed.

zunath:
﻿@coderabbitai review

For context: the one actionable finding from the previous review (hardcoded 30-minute delay) was fixed in 5dcc215b3, and 625b89736 merged the latest `feature/combat-upgrade`. Neither commit has been reviewed yet, since the last run only covered up to 223e7c107.

PR #2113, opened 2026-07-21
Fix stealth exit on detection

zunath:
## Summary

- Exit a player's stealth mode when an opposed Spot check successfully detects them.
- Notify the detected player while preserving observer-specific stealth behavior for NPCs and DMs.
- Add regression coverage and document the player reveal behavior.

## Root cause

The custom Spot detection handler returned a successful detection result to the engine but never cleared the target player's `ActionMode.Stealth`. The player therefore remained in stealth and kept the associated status and stamina drain active after being detected.

## Player impact

Detected players now leave stealth immediately and the existing stealth-exit handler removes the status effect, clears cached verdicts, and stops stamina drain. NPC encounter behavior is unchanged.

## Validation

- `dotnet build SWLOR.Game.Server.Tests\SWLOR.Game.Server.Tests.csproj -p:RunPostBuildEvent=Never`
- `dotnet test SWLOR.Game.Server.Tests\SWLOR.Game.Server.Tests.csproj --no-build --filter "FullyQualifiedName~EspionageSystemTests"` — 12 passed

### DIFF
--- a/SWLOR.Game.Server.Tests/Perks/EspionageSystemTests.cs
+++ b/SWLOR.Game.Server.Tests/Perks/EspionageSystemTests.cs
@@ -114,6 +114,23 @@ public class EspionageSystemTests
     }
 
     [Test]
+    public void SuccessfulSpotDetection_ExitsPlayerStealth()
+    {
+        var stealthSource = File.ReadAllText(Path.Combine(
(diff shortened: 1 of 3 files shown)


zunath:
@codex review

PR #2115, opened 2026-07-21
Fix transient module packing failures

zunath:
## Summary

- Retry individual `nwn_gff` resource conversions up to three times with short backoff.
- Remove partial conversion output before retrying and retain the final converter error when retries are exhausted.
- Skip the final key prompt for redirected input while preserving the interactive pause.
- Update the module-packer documentation and republish the tracked CLI executable.

## Root cause

`nwn_gff.exe` intermittently exits with code 1 during sustained parallel module packing even when the resource is valid. The initially reported `imp_poly_3.uti.json` converts successfully by itself, and repeated full packs surfaced first-attempt failures on different unrelated resources.

## Impact

Transient native converter failures no longer discard an otherwise healthy multi-minute pack. Persistently invalid resources still fail the operation after three attempts. Noninteractive invocations also finish cleanly after successful module assembly.

## Validation

- `dotnet build SWLOR.CLI\SWLOR.CLI.csproj -p:RunPostBuildEvent=Never`
- Published the self-contained CLI using `FolderProfile`
- Completed a full pack of 19,173 resources and 310 scripts with exit code 0
- Confirmed multiple transient first-attempt failures recovered automatically

### DIFF
--- a/SWLOR.CLI/ModulePacker.cs
+++ b/SWLOR.CLI/ModulePacker.cs
@@ -15,6 +15,8 @@ namespace SWLOR.CLI
         private const int MaxDefaultWorkerCount = 12;
         private const int ProgressInterval = 100;
         private const int ReservedProcessorCount = 2;
+        private const int ResourceConversionRetryCount = 3;
+        private const int ResourceConversionRetryDelayMilliseconds = 250;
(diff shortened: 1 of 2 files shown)


zunath:
@codex review

PR #2130, opened 2026-07-22
Prevent Guard procs on discarded attacks

zunath:
## Summary

- thread the native attack result from `GetDamageRoll` into Guard resolution
- reject discarded, plot-target, and clearly non-hostile NPC swings before rolling Guard
- add regression coverage for the landed-hit gate and native attack-result plumbing

## Root cause

The landed-attack gate from an earlier Guard fix was lost during a later combat-service restoration. Neverwinter Nights invokes the native damage-roll hook for some swings it later discards. Because Guard was again checking only calculated positive physical damage, those discarded attacks could trigger Guard feedback, Counter Ready, Retaliation Pulse, recovery, and enmity while the player appeared idle or far from an enemy.

## Player impact

Guard and its riders now activate only for actual landed hostile physical attacks. Intended Guard mitigation, retaliation damage, and enmity behavior are unchanged.

## Validation

- `dotnet build SWLOR.Game.Server.Tests\SWLOR.Game.Server.Tests.csproj -p:RunPostBuildEvent=Never` — passed
- `dotnet test SWLOR.Game.Server.Tests\SWLOR.Game.Server.Tests.csproj --no-build --filter "FullyQualifiedName~CombatDamageTests.GuardedHitModifiers_OnlyRunForPhysicalDamageFromDamageRoll"` — passed
- broader `CombatDamageTests` run: 47 passed; 3 unrelated asset tests could not run because this worktree's `SWLOR_Haks` submodule content is not checked out

## Summary by CodeRabbit

- **Bug Fixes**
  - Guard mitigation and related effects now trigger only when a physical attack successfully lands on a valid target.
  - Discarded, unsuccessful, non-physical, and invalid-target attacks no longer incorrectly activate guarded-hit mechanics.
  - Non-hostile NPC attacks are excluded from guard processing, while supported player- and DM-driven combat scenarios continue to work correctly.

### DIFF
--- a/SWLOR.Game.Server.Tests/Service/CombatDamageTests.cs
+++ b/SWLOR.Game.Server.Tests/Service/CombatDamageTests.cs
@@ -555,6 +555,11 @@ public class CombatDamageTests
         guard.IndexOf("!isLandedAttack", StringComparison.Ordinal).Should().BeLessThan(
             guard.IndexOf("var guardChance", StringComparison.Ordinal),
             "discarded swings must never enter the Guard roll");
+        var guardableSource = ExtractMethod(combatSource, "private static bool IsGuardableAttackSource(");
+        guardableSource.Should().Contain("GetIsReactionTypeHostile(attacker, defender)");
+        guardableSource.Should().Contain("GetIsReactionTypeHostile(defender, attacker)");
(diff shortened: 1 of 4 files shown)


zunath:
@codex review

PR #2131, opened 2026-07-23
Fix cross-platform Docker Compose compatibility

zunath:
## Summary

- Replace `${PWD-.}` interpolation with Compose-relative paths.
- Apply portable paths consistently to both environment files and all bind mounts.
- Restore the file's historical `version: "3.7"` declaration for legacy `docker-compose` compatibility.
- Preserve the existing service configuration and container paths.

## Root cause

The Linux deployment uses a legacy Compose parser. It rejected `${PWD-.}` interpolation, and without a top-level schema version it interpreted the file as legacy v1 format. That caused `services` and `volumes` to be treated as service names and produced `Unsupported config option` errors.

Compose-relative `./...` paths avoid host-shell interpolation. Restoring the repository's prior Compose 3.7 declaration tells legacy Compose to parse the `services` and top-level `volumes` sections correctly.

## Impact

The same Compose file can be parsed by legacy Linux `docker-compose` and modern Docker Desktop Compose on Windows. Modern Compose accepts the version declaration but emits a non-fatal warning that the field is obsolete.

## Validation

- Windows: `docker-compose -f SWLOR.Game.Server/Docker/docker-compose.yml config --quiet`
- Ubuntu/WSL2: `docker-compose config --quiet`
- `git diff --check`

### DIFF
--- a/SWLOR.Game.Server/Docker/docker-compose.yml
+++ b/SWLOR.Game.Server/Docker/docker-compose.yml
@@ -3,7 +3,7 @@ services:
         hostname: redis
         image: redislabs/redismod:latest
         volumes:
-          - ${PWD-.}/redis:/data
+          - ./redis:/data
         restart: unless-stopped
                 
     redis-commander:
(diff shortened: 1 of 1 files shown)


zunath:
@codex review

PR #2147, opened 2026-08-04
Fix Toolset Go To object navigation

zunath:
## Summary

- make Source-tab **Go To** activate the target area and bring **Area Contents** to the front
- select the exact placed instance in both the area editor and Area Contents tree
- expand the target's tree ancestors, clear a hiding filter, and automatically scroll the row into view
- retain camera and tree navigation requests while a large area or its views are still loading
- zoom the viewport from full-area framing to the selected object
- realize explicitly requested instances beyond the normal 200-row large-group cap without removing that performance safeguard

## Root cause

Go To activated the area document and selected its underlying instance row, but the camera request was transient and could fire before the viewport existed. Area Contents was not explicitly activated and had no corresponding reveal/scroll request, so its grouped tree could remain collapsed or omit a target beyond the realization cap.

## Impact

Go To now completes the full navigation action for both cold and already-open areas: the correct area opens, the object is selected, the viewport zooms to it, Area Contents selects the same instance, and the scrollbar follows it. Switching tabs while a large area loads no longer loses the pending focus.

## Validation

- `dotnet build SWLOR.Toolset.Tests\SWLOR.Toolset.Tests.csproj --no-restore -p:RunPostBuildEvent=Never`
- focused Go To, Area Contents, viewport, source, merchant, and context-menu suite: 65 passed
- `git diff --check`

## Summary by CodeRabbit

* **New Features**
  * Revealing area placements now automatically opens the Area Contents panel, expands the relevant hierarchy, selects the placement, and scrolls it into view.
  * Camera focus and placement navigation requests are preserved while scenes, tabs, or panels are loading.
  * Large placement groups can now reveal items beyond the initial display limit.

* **Bug Fixes**
  * Improved camera focus reliability during scene loading and tab changes.
  * Prevented filters and delayed panel attachment from blocking placement navigation.

### DIFF
--- a/SWLOR.Toolset.Tests/AreaContentsTests.cs
+++ b/SWLOR.Toolset.Tests/AreaContentsTests.cs
@@ -1,5 +1,9 @@
 using System.Numerics;
 using System.Text;
+using Avalonia.Controls;
+using Avalonia.Headless.NUnit;
+using Avalonia.Threading;
+using Avalonia.VisualTree;
 using FluentAssertions;
 using NUnit.Framework;
 using SWLOR.NWN.Formats.Common;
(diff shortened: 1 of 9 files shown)


zunath:
@codex review

PR #2171, opened 2026-08-17
Fix ranged NPC recovery after area abilities

zunath:
## Summary

- evaluate self-centered hostile AOEs around the caster using the ability's actual targeting radius
- preserve a valid pre-cast target when NPC enmity is transiently empty
- allow autonomous NPCs to reacquire targets after delayed ability recovery
- resume attacks through the existing ranged stand-off logic so rifle and pistol users do not run into melee

## Root cause

All hostile area abilities used target-centered cluster selection, including abilities whose area originates on the caster. Ranged NPCs could therefore choose an AOE because enemies were clustered around their current target even when nobody was inside the caster-centered effect. Afterward, recovery could discard a valid saved target when enmity was briefly empty, leaving the NPC idle or falling back into bad movement behavior.

## Validation

- `dotnet build SWLOR.Game.Server.Tests\SWLOR.Game.Server.Tests.csproj -p:RunPostBuildEvent=Never`
- `dotnet test SWLOR.Game.Server.Tests\SWLOR.Game.Server.Tests.csproj --no-build --filter "FullyQualifiedName~AIModelTests|FullyQualifiedName~AIProfileValidationTests|FullyQualifiedName~AbilityDamageQueueTests|FullyQualifiedName~EnmityTests"` (87 passed)

No HAK submodule files changed.

### DIFF
--- a/SWLOR.Game.Server.Tests/AI/AIModelTests.cs
+++ b/SWLOR.Game.Server.Tests/AI/AIModelTests.cs
@@ -270,6 +270,27 @@ public class AIModelTests
     }
 
     [Test]
+    public void AITarget_SelfCenteredHostileAreasEvaluateEnemiesAroundTheCaster()
+    {
+        var source = ReadSource("SWLOR.Game.Server", "Service", "AIService", "AITarget.cs")
+            .Replace("\r\n", "\n");
+        var selfCenteredBody = source.Substring(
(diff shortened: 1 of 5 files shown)


zunath:
Superseded because this PR reused a previously merged branch. Reopening the same fix from a clean branch based on the current `feature/combat-upgrade` head.

